PR #17, opened 2019-11-14
Bug/do experiments print output

duylethanh:
* Fixed a bug where users would be instructed to use additional cli arguments even though they were already provided
* Added informative print to do_experiments_purge when user does not provide any additional arguments

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -404,10 +404,12 @@ experiments_purge_parser.add_argument('-f', action='store_true')
 def do_experiments_print_output(args):
 	cfg = extl.base.config_for_dir()
 
-	selection = list(select_runs_from_cli(cfg, args, default_all=False))
-
-	if selection:
-		for run in selection:
+	if not can_select_runs_from_cli(args):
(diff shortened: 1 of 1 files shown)


scripts/simex
```diff
@@ -362,39 +362,43 @@ experiments_launch_parser.add_argument('--queue', type=str)
...
-			except FileNotFoundError:
-				pass
-			try:
-				os.unlink(run.output_file_path('status.tmp'))
-			except FileNotFoundError:
-				pass
-		else:
-			print("This would purge experiment '{}', instance '{}' [{}]".format(
-					exp.name, instance, run.repetition))
+	if can_select_runs_from_cli(args):
```
avdgrinten: This would be much prettier if we just did:
```
if not can_select_runs_from_cli(args):
    print(...)
    return
```

duylethanh: I have made the changes and also added a small help message for the '-f' option when calling 'simex e purge -h'.

PR #20, opened 2019-11-29
Feature/additional slurm args

duylethanh:
This PR adds functionality to support user-defined slurm parameters in the experiments.yml for each experiment. If the user provides parameters that simexpal sets, e.g. the Jobname or the output file, the user arguments will be prioritized.

I could change it, so that simexpal's parameters get priority in case we need that.


avdgrinten:
This PR can now be rebased to remove the unnecessary commit.

duylethanh:
> This PR can now be rebased to remove the unnecessary commit.

I removed the unnecessary commit. 

Furthermore I added a commit where I replaced some lines in the cli by using the new 'display_name' property.

PR #21, opened 2019-12-06
Feature/print error output

duylethanh:
* 'do_experiments_purge()' now also deletes the respective .stderr file
* 'do_experiments_print_output()' now prints both the output and error output file by default
* Refactoring of 'do_experiments_list()' using a new IntEnum class for status codes
* Refactoring of 'select_runs_from_cli()'

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -1,5 +1,6 @@
 
 from collections import OrderedDict
+from enum import IntEnum
 import copy
 import itertools
 import os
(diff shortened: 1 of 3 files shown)


simexpal/base.py
```diff
@@ -725,6 +766,33 @@ def output_file_path(self, ext):
 		return os.path.join(self.experiment.output_subdir,
 				get_output_file_name(ext, self.instance.filename, self.repetition))
 
+	def get_status(self):
+		in_submission = os.access(self.aux_file_path('lock'), os.F_OK)
+		submitted = os.access(self.aux_file_path('run'), os.F_OK)
+		started = os.access(self.output_file_path('out'), os.F_OK)
+		finished = os.access(self.output_file_path('status'), os.F_OK)
```
avdgrinten: Instead of always accessing all files, we should access them only when necessary (for performance reasons: `os.access` needs to perform filesystem I/O and is relatively expensive).

duylethanh: Done.

PR #23, opened 2020-01-09
Feature/multiple input files

duylethanh:
* Refactoring of 'Instance.filename' with 'Instance_get_unique_name()' and marking 'Instance.filename' as deprecated
* Added support for experiments with multiple input files

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -442,6 +442,12 @@ class Instance:
 		else:
 			return [self.yml_name]
 
+	@property
+	def unique_filename(self):
+		if len(self.filenames) > 1:
+			raise RuntimeError("The instance '{}' does not have a unique filename.".format(self.yml_name))
+		return self.filenames[0]
+
 	def check_available(self):
 		for file in self.filenames:
 			if not os.path.isfile(os.path.join(self._cfg.instance_dir(), file)):
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -123,8 +123,8 @@ class Config:
 		def construct_instances():
 			if 'instances' in self.yml:
 				for inst_yml in self.yml['instances']:
-					for item in inst_yml['items']:
-						yield Instance(self, item, inst_yml)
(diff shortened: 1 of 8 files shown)


simexpal/base.py
```diff
@@ -122,17 +122,17 @@ def __init__(self, basedir, yml):
...
 		def construct_variants():
 			if 'variants' in self.yml:
 				for axis_yml in self.yml['variants']:
 					for variant_yml in axis_yml['items']:
 						yield Variant(self, axis_yml['axis'], variant_yml)
 
-		for inst in sorted(construct_instances(), key=lambda inst: inst.filename):
-			self._insts[inst.filename] = inst
+		for inst in sorted(construct_instances(), key=lambda inst: inst.get_unique_name()):
+			self._insts[inst.get_unique_name()] = inst
```
avdgrinten: How does this work if the instance does not have a unique file?

simexpal/base.py
```diff
@@ -375,22 +375,56 @@ def collect_successful_results(self, parse_fn):
...
+		warnings.warn("The 'Instance.filename' attribute is deprecated and will be removed in future versions. "
+						"Use 'Instance.get_unique_name()' instead.", DeprecationWarning)
+
+		return self.get_unique_name()
+
+	def get_unique_name(self):
+		if isinstance(self._inst_yml['items'][self.index], dict):
+			assert 'name' in self._inst_yml['items'][self.index]
+
+			return self._inst_yml['items'][self.index]['name']
```
avdgrinten: This does not return the file name but the instance name (?).

duylethanh:
I made the changes as we have discussed (replacing `Instance.filename` appropriately with `Instance.shortname` or `Instance.unique_filename`; check for unique instance, build, revision, variant and experiment names)
Furthermore the `Instance.yml_name` property contains the name of the instance as specified in the `experiments.yml`. The property name might be subject to change.

PR #24, opened 2020-01-14
Stdout capture

duylethanh:


simexpal/launch/common.py
```diff
@@ -218,12 +218,19 @@ def compile_manifest(run):
...
-	with open(manifest.output_file_path('out'), "w") as f:
-		# We do not actually need to write anything to the output file.
-		# However, we might want to pipe experimental output to it.
-		if manifest.output == 'stdout':
+	(stdout_pipe, stdout) = (None, None)
+	if manifest.output == 'stdout':
+		with open(manifest.output_file_path('out'), "w") as f:
+			# We do not actually need to write anything to the output file.
+			# However, we might want to pipe experimental output to it.
 			stdout = os.dup(f.fileno())
```
avdgrinten: We should still create the output file even if `output != 'stdout'`. In that case, however, we don't need to `dup` the file handle.

examples/sorting/experiments.yml
```diff
@@ -15,7 +15,5 @@ instances:
 experiments:
   - name: insertion-sort
     args: ['./sort.py', '--algo=insertion-sort', '@INSTANCE@']
-    output: stdout
```
avdgrinten: We should not change the example here (its output is needed to make an evaluation script work).

duylethanh: I reverted the changes to the `experiment.yml`.

simexpal/launch/common.py
duylethanh: Done. The check `manifest.output == 'stdout'` is done after creating the output file.

PR #26, opened 2020-02-11
docs

duylethanh:
This PR contains C++ code for Quicksort with base blocks and documentation for automated builds and the run matrix.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -295,6 +295,7 @@ simexpal will duplicate the experiment for each possible combination
 of variants. Such a combination will consist of exactly one variant
 for every ``axis`` property.
 
+
 Automated Builds and Revision Support
 -------------------------------------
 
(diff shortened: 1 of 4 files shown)


docs/user_guide.rst
```diff
@@ -307,49 +310,117 @@ in ``experiments.yml``.
...
+            - '-j@PARALLELISM@'
+        install:
+          - args:
+            - 'make'
+            - 'install'
+
+    revisions:
+      - name: main
+        build_version:
+          'simexpal': 'master'      # specify a branch or a tagged commit
```
avdgrinten: In principle, specifying a SHA1 hash is the best way to ensure reproducibility.

examples/sorting_cpp/quickSort.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include <string>
+
+int block_size = 16;
+void (*block_algo)(std::vector<int>&, int, int);
+
+void swap(int *a,int *b){
+    int t = *a;
+    *a = *b;
+    *b = t;
+}
```
avdgrinten: `std::swap(*a, *b)` can be used instead.

examples/sorting_cpp/quickSort.cpp
```diff
@@ -0,0 +1,127 @@
+#include <iostream>
+#include <vector>
+#include <fstream>
+#include <string>
+
+int block_size = 16;
+void (*block_algo)(std::vector<int>&, int, int);
```
avdgrinten: Naming is inconsistent here. I think we should use either `camelCase` or `snake_case` everywhere.

avdgrinten:
Thank you for the PR. LGTM.

PR #29, opened 2020-02-26
command line parameter to pick different variants

tzovas:
Option to pick variants and variant combination by name


duylethanh:
Feature was merged with PR #47.

PR #30, opened 2020-03-03
Print exp info

tzovas:
Print information about the chosen experiments


duylethanh:
Feature was merged with PR #47.

PR #31, opened 2020-03-10
Support development builds

duylethanh:
* support for dev-build revisions which allow running experiments with programs currently in development (without pushing changes and changing the respective commit hashes in the `experiments.yml`)
* some refactoring regarding status queries of builds and the construction of the `Config._revisions` dictionary
* simexpal now reserves names starting with an underscore (e.g `_dev`). An error is thrown if such names are used in the `experiments.yml`

A typical workflow using dev-builds could look like this:
1. create a revision with the normal syntax
2. add `develop: True` to a revision and remove the commit-hashes (but leave an empty string i.e. `''`, simexpal will pull the latest commit for dev-build revision)
3. use `simex develop <build> --revision <revision_name>` to build the dev-build revision
4. make changes to the program source code in the directory `develop/<build>@<revision_name>.src`
5. use `simex develop <build> --revision <revision_name>` to rebuild it 
6. run experiments
7. repeat steps 3-6 if necessary

It is also possible to remove the `name` attribute for **one** dev-build. Internally simexpal will set the name of this revision to `_dev`.
In most cases there will only be one dev-build and this allows syntactic sugar when working with dev-builds. For example the command `simexpal develop <build>` will do the same as the command `simexpal develop <build> --revision _dev`.
Also for this default dev-build the directories do not have the `@<revision>` suffix.

I will create a PR with a more detailed description of this feature in the future.

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -96,6 +96,70 @@ run_selection_parser.add_argument('--failed', action='store_true')
 run_selection_parser.add_argument('--unfinished', action='store_true')
 
 # ---------------------------------------------------------------------------------------
+
+def select_phases_from_cli(args):
+
+	def subsequent_phases(start_phase):
+		return [phase for phase in simexpal.build.Phase if phase >= start_phase]
+
+	if args.recheckout:
+		if not args.f:
+			print("This would delete the local git repository for the build and reclone it. Confirm this action"
+					" by using the '-f' flag.")
+			return []
+		else:
+			return subsequent_phases(simexpal.build.Phase.CHECKOUT)
+	elif args.checkout:
+		if not args.f:
+			print("This would delete the local git repository for the build and reclone it. Confirm this action"
+					" by using the '-f' flag.")
+			return []
+		else:
+			return [simexpal.build.Phase.CHECKOUT]
+	elif args.reregenerate:
+		return subsequent_phases(simexpal.build.Phase.REGENERATE)
+	elif args.regenerate:
+		return [simexpal.build.Phase.REGENERATE]
+	elif args.reconfigure:
+		return subsequent_phases(simexpal.build.Phase.CONFIGURE)
+	elif args.configure:
+		return [simexpal.build.Phase.CONFIGURE]
+	elif args.compile:
+		return [simexpal.build.Phase.COMPILE]
+	elif args.reinstall:
+		return subsequent_phases(simexpal.build.Phase.INSTALL)
+	elif args.install:
+		return [simexpal.build.Phase.INSTALL]
+	else:
+		# This case also includes 'args.recompile == True'
+		return subsequent_phases(simexpal.build.Phase.COMPILE)
+
+phase_selection_parser = argparse.ArgumentParser(add_help=False)
+phase_selection_mechanism = phase_selection_parser.add_mutually_exclusive_group()
+phase_selection_mechanism.add_argument('--recheckout', action='store_true',
+			help='Delete local build git repository, reclone, regenerate, reconfigure, recompile and reinstall it')
+phase_selection_mechanism.add_argument('--checkout', action='store_true',
+			help='Delete local git repository for build and (re-)clone it')
+phase_selection_mechanism.add_argument('--reregenerate', action='store_true',
+			help='Regenerate, reconfigure, recompile and reinstall build')
+phase_selection_mechanism.add_argument('--regenerate', action='store_true',
+			help='(Re-)Regenerate build')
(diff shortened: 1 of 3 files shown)


simexpal/base.py
```diff
@@ -120,6 +120,12 @@ def __init__(self, basedir, yml):
 		self._variants = OrderedDict()
 		self._exp_infos = OrderedDict()
 
+		def check_key_value(yml_dict, target_key):
```
avdgrinten: I think this is rather inelegant: we assume here that `yml_dict` has a key containing the name. It's not clear to me that this is general enough in all cases.

avdgrinten: Also, the name of the function is not really self-explanatory. What about something like: `check_for_reserved_name()`?

simexpal/base.py
```diff
@@ -134,7 +134,11 @@ def construct_instances():
 
 		def construct_variants():
 			if 'variants' in self.yml:
+				check_key_value(self.yml['variants'], 'axis')
```
avdgrinten: Since we know the key here, can't we simply pass `self.yml['variants']['axis']`? (Same for the other cases below.)

scripts/simex
```diff
@@ -253,6 +253,37 @@ builds_make_parser.set_defaults(cmd=do_builds_make)
 
 # ---------------------------------------------------------------------------------------
 
+def do_dev_builds(args):
```
avdgrinten: Rename to `do_develop` (for consistency with other commands).

scripts/simex
```diff
@@ -253,6 +253,37 @@ builds_make_parser.set_defaults(cmd=do_builds_make)
...
+	return do_dev_builds_make(args)
+
+dev_builds_parser = main_subcmds.add_parser('develop', help='Build local programs',
+		aliases=['d'], parents=[phase_selection_parser])
+dev_builds_parser.set_defaults(cmd=do_dev_builds)
+dev_builds_parser.add_argument('--revision', type=str)
+dev_builds_parser.add_argument('build', type=str)
+
+def do_dev_builds_make(args):
```
avdgrinten: Put this directly into `do_develop`.

scripts/simex
```diff
@@ -253,6 +253,37 @@ builds_make_parser.set_defaults(cmd=do_builds_make)
...
 # ---------------------------------------------------------------------------------------
 
+def do_dev_builds(args):
+	return do_dev_builds_make(args)
+
+dev_builds_parser = main_subcmds.add_parser('develop', help='Build local programs',
+		aliases=['d'], parents=[phase_selection_parser])
+dev_builds_parser.set_defaults(cmd=do_dev_builds)
+dev_builds_parser.add_argument('--revision', type=str)
+dev_builds_parser.add_argument('build', type=str)
```
avdgrinten: We could support multiple builds on the command line.

simexpal/base.py
```diff
@@ -657,6 +657,16 @@ def prefix_dir(self):
...
+	@property
+	def source_dir(self):
+		"""
+			dev-builds only have a source directory instead of a repo and clone directory
+		"""
+		assert self.revision.is_dev_build
+
+		rev = self._get_dev_build_suffix()
+		return os.path.join(self._cfg.basedir, 'develop', self.name + rev + '.src')
+
```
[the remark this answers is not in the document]

simexpal/base.py
duylethanh: > Also, the name of the function is not really self-explanatory. What about something like: `check_for_reserved_name()`?

seems like a good name, done

duylethanh: > I think this is rather inelegant: we assume here that `yml_dict` has a key containing the name. It's not clear to me that this is general enough in all cases.

I changed `check_for_reserved_name()` to only take a name as input and also added a print of the name if it does not pass the check

simexpal/base.py
duylethanh: done (see remark above)

scripts/simex
duylethanh: done

scripts/simex
duylethanh: done

scripts/simex
duylethanh: I renamed the `build` argument to `builds` and added the parameter `nargs='+'` to the argument. You can now use the command `simex develop <build1> <build2> <build ...>` to apply actions on `build1`, `build2`, ...

duylethanh:
> Looks mostly good: requests for some minor changes.

I made all the requested changes. The PR should be ready to merge

avdgrinten:
LGTM. Thank you!

PR #34, opened 2020-03-13
Docs

duylethanh:
This PR contains documentation for instances in the `experiments.yml`.

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -1,3 +1,5 @@
+.. _CommandLineReference:
+
 Command-line Reference
 ======================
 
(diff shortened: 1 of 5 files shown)


avdgrinten:
Thanks for the PR. The text looks good but I'm not sure if the experiment.yml reference page is the best location for it. I'd like to see a reference of the experiments.yml file sorted by keys/blocks.

We should still keep the text that you wrote, just move it somewhere else. Possibly to an own "Instances" page. We could rename the `Usage Guide` to `Quick Start` or something similar to reflect this.

duylethanh:
> I'd like to see a reference of the experiments.yml file sorted by keys/blocks.

Is it not already (see image below) or do you mean something else?

image


> We could rename the `Usage Guide` to `Quick Start` or something similar to reflect this.

This sounds like a good idea.

duylethanh:
I updated the PR. I moved the text to an own Instance page and made slight changes to it. I also added a reference for the keys used for instances in the `experiments.yml` and renamed `Usage Guide` to `Quick Start`.

avdgrinten:
Thank you for the PR! Merging...

PR #36, opened 2020-03-20
Bugfix for build process and removal of unused variables

duylethanh:
* small bug fix regarding the phase handling during the build process
* ~~prefix directories are appended to the PATH environment variable~~
* removal of unused/unnecessary variables
* added `default_workdir` keyword in `do_step()` call in the regenerate phase to make the code consistent with the other phases

### DIFF
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -38,7 +38,7 @@ def lock_run(run):
 def create_run_file(run):
 
 	# Create the .run file. This signals that the run has been submitted.
-	with open(run.aux_file_path('run.tmp'), "w") as f:
+	with open(run.aux_file_path('run.tmp'), "w"):
 		pass
 	os.rename(run.aux_file_path('run.tmp'), run.aux_file_path('run'))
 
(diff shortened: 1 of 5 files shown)


avdgrinten:
This PR needs a more specific title.

simexpal/launch/common.py
```diff
@@ -139,6 +139,7 @@ def get_paths(self):
 		paths = []
 		for build_yml in self.yml['builds']:
 			paths.append(os.path.join(build_yml['prefix'], 'bin'))
+			paths.append(build_yml['prefix'])
```
avdgrinten, line 142: Why this change? That seems strange to me.

duylethanh: There were cases where users had their executables directly in the prefix-dir.

avdgrinten: But isn't the fix to move them into `bin/` then? Or, we could add a `extra_paths` argument to allow the the user to configure this behavior. In any case, adding the prefix directory unconditionally does not sound like a good resolution to me.

duylethanh: Ok, that sounds reasonable. I will remove this line and create an issue regarding the `extra_paths` argument so that I do not forget about it and implement it in the future.

duylethanh:
PR is ready to be merged.

avdgrinten:
Thank you!

PR #40, opened 2020-04-03
@-Variables for build-dirs, bugfix and minor refactor

duylethanh:

- Added support for `@THIS_COMPILE_DIR@`, `@SOURCE_DIR_FOR:<build_name>@`, `@SOURCE_DIR_FOR:<build_name>@` and `@SOURCE_DIR_FOR:<build_name>@` variable in build args
- minor refactoring of string matching in `substitute()` in `invoke_run()`
- fix for a bug where `'_none'` was displayed as revision in the terminal for experiments that do not have a revision
- remove usage of Revision objects that had `'_none'` as a name by adapting the sorting key that is passed into `Config._expand_matrix()`

### DIFF
--- a/simexpal/build.py
+++ b/simexpal/build.py
@@ -96,6 +96,8 @@ def make_build_in_order(cfg, build, wanted_builds, wanted_phases):
 				return build.clone_dir
 			else:
 				return build.source_dir
+		elif var == 'THIS_COMPILE_DIR':
+			return build.compile_dir
 		elif var == 'THIS_PREFIX_DIR':
 			return build.prefix_dir
 		elif var == 'PARALLELISM':
--- a/simexpal/build.py
+++ b/simexpal/build.py
@@ -99,6 +99,10 @@ def make_build_in_order(cfg, build, wanted_builds, wanted_phases):
 		other_build = cfg.get_build(build_name, build.revision)
 		return other_build.compile_dir
 
+	def get_prefix_dir_for(build_name):
+		other_build = cfg.get_build(build_name, build.revision)
+		return other_build.prefix_dir
+
 	def substitute(var):
 		# 'THIS_SOURCE_DIR' is prefered, 'THIS_CLONE_DIR' is deprecated
 		if var in ['THIS_CLONE_DIR', 'THIS_SOURCE_DIR']:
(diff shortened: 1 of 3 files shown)


simexpal/build.py
```diff
@@ -89,6 +89,17 @@ def get_concurrency():
 			n = os.cpu_count()
 		return n
 
+	def get_source_dir_for(build_name):
+		if build_name == build.name:
+			if not build.revision.is_dev_build:
+				return build.clone_dir
+			return build.source_dir
+
+		other_build = cfg.get_build(build_name, build.revision)
```
avdgrinten, line 98: Why do we special case the current build here?

simexpal/build.py
```diff
@@ -100,6 +100,13 @@ def get_source_dir_for(build_name):
 			return other_build.clone_dir
 		return other_build.source_dir
 
+	def get_compile_dir_for(build_name):
+		if build_name == build.name:
+			return build.compile_dir
+
+		other_build = cfg.get_build(build_name, build.revision)
```
avdgrinten, line 107: Same here.

simexpal/build.py
```diff
@@ -107,6 +107,13 @@ def get_compile_dir_for(build_name):
 		other_build = cfg.get_build(build_name, build.revision)
 		return other_build.compile_dir
 
+	def get_prefix_dir_for(build_name):
+		if build_name == build.name:
+			return build.prefix_dir
+
+		other_build = cfg.get_build(build_name, build.revision)
```
avdgrinten, line 114: And here.

simexpal/build.py
duylethanh: It saves us a call of `cfg.get_build()`, which loops over all builds and revision. I can remove this case if you think it is overkill

avdgrinten: That sounds more like an efficiency problem in `cfg.get_build()` to me. If this turns out to be too slow, we could always add a `dict()` of all builds.

Yes, please remove this special case here to simplify the code.

duylethanh: Done.

duylethanh:
I made the changes. The PR is ready to be merged.

avdgrinten:
Thank you for the PR.

Issue #39, opened 2020-03-30
Compact experiments list


angriman:
When dealing with many variations the full list of experiments printed by `simex e` might be too long to visualize. It would be nice to have a "compact view" of experiments that shows for each experiment the status of each variation e.g.:
```
$ simex e --compact
experiment 1: 0/100 finished, 50/100 submitted, 50/100 started
experiment 2: ...
...
```

Does this sound reasonable?

avdgrinten:
Yes, sounds like a good idea. We had something similar planned already but so far, no code has been written yet.

duylethanh:
@avdgrinten how "detailed" should the print-out be? If we choose to do it like @angriman proposed I would probably format it like
```
$ simex e --compact
experiment      not submitted    submitted    started    finished   failed
-----------      --------------   -----------   --------   -------- -------
experiment1    0/100                  5/100            75/100    10/100    10/100
experiment2     ...
```
(The entries are shifted  due to some rendering issue, but you get the gist)

avdgrinten:
I am not sure if we need columns for "submitted" and "not submitted" experiments. We could have an extra column "other" with text like "5/100 submitted, 1/100 not submitted" (where only the non-zero entries are shown).

angriman:
Wouldn't that be simpler to just have a "submitted" column? An experiment is either submitted or not right?

avdgrinten:
This has been fixed in the meantime.

PR #42, opened 2020-04-14
compact experiments list and full experiment name

duylethanh:
- refactoring of `do_experiments_list`
- addition of constant `EXPERIMENTS_LIST_THRESHOLD=30`
- implementation of https://github.com/hu-macsy/simexpal/issues/39
- added `--full` as option for `simex experiments list`, which will set the experiment column width to the length of the longest experiment `name`/`display_name` instead of using a fixed value

Calling `simex experiments` will now display a **detailed** list of runs if the total number of runs is 
`< EXPERIMENTS_LIST_THRESHOLD` and a **compact** list otherwise.

Calling `simex experiments list --compact` will display a compact list of all runs and looks like this:
simex_experiments_compact

Calling `simex experiments list --detailed` will display a detailed list of all runs (how it is displayed currently).

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -286,6 +286,7 @@ dev_builds_parser.add_argument('builds', nargs='+', type=str)
 def do_experiments(args):
 	args.detailed = False
 	args.compact = False
+	args.full = False
 
 	return do_experiments_list(args, as_default_subcmd=True)
 
@@ -305,7 +306,7 @@ def do_experiments_list(args, as_default_subcmd=False):
 			return colors['red']
 		return ''
 
-	def show_compact_list():
+	def show_compact_list(calc_exp_len=False):
 
 		def print_experiment_statistics():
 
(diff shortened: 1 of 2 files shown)


scripts/simex
```diff
@@ -318,31 +318,32 @@ def do_experiments_list(args, as_default_subcmd=False):
...
-																			f_entry,
-																			o_entry,
+													include_status_string=True)
+
+			print('{:{len}.{len}} {}{:10.10}{} {}{:10.10}{} {}{:20.20}{} {}{}{}'.format(exp_name,
+																						*started_statistics[0],
+																						*finished_statistics[0],
+																						*failures_statistics[0],
+																						*other_statistics[0],
+																						len=exp_len))
```
avdgrinten, line 341: Indentation is broken here.

scripts/simex
```diff
@@ -300,6 +305,68 @@ def do_experiments_list(args, as_default_subcmd=False):
...
+						prefix = ''
+						if include_status_string:
+							prefix = str(s) + ': '
+
+						entry_list.append(prefix + str(status_dict[s]) + '/' + str(num_runs))
+
+				if len(entry_list) == 0:
+					return entry_list.append('')
+
+				return entry_list
```
avdgrinten: This is a bit ugly, in particular since we return an array and often only need one argument. Why don't you let this function act only on a single status and call it multiple times below (e.g. inside a list comprehension) when you need the result on multiple statuses?

avdgrinten:
Looks mostly good. Can be merged once the two issues above are resolved.

scripts/simex
duylethanh: Done

duylethanh:
> Calling `simex experiments --compact` will display a compact list of all runs and looks like this:

I forgot the keyword `list`, i.e. calling `simex experiments list --compact` will display a compact list of all runs.
> Calling `simex experiments --detailed` will display a detailed list of all runs (how it is displayed currently).

same here: calling `simex experiments list --detailed` will display a detailed list of all runs.

I edited the comment above.

scripts/simex
duylethanh: As discussed we will leave the code as it is for now, since it easily handles zero-entries.
I only removed the `if len(entry_list) == 0` case as it is handled by the `fillvalue` in `zip_longest`.

duylethanh:
I made the changes. The PR is ready to be merged

PR #45, opened 2020-05-11
Update README.md

cdeschryver:
current link is broken, set it to the quick start guide


avdgrinten:
Good catch, thank you for the PR!

PR #46, opened 2020-05-11
Added [at]-variable for output subdirectory and enables it to be repl…

AndreChinazzo:
…aced in workdir


avdgrinten:
LGTM, thank you for the PR.

PR #50, opened 2020-05-17
Fix #48

### LINKED ISSUE #48: Grouping of experiments by Slurm launcher is broken
The Slurm launcher used to group jobs for the same experiment/revision/variation tuple (e.g. jobs that only differ in the instance) into an array job. This does not work anymore:
```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
            100996      core bghkkn ~ grintena PD       0:00      1 (Resources)
            100997      core bghkkn ~ grintena PD       0:00      1 (Priority)
            100998      core bghkkn ~ grintena PD       0:00      1 (Priority)
            100999      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101000      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101001      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101002      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101003      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101004      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101005      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101006      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101007      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101008      core bghkkn ~ grintena PD       0:00      1 (Priority)
            101009      core bghkkn ~ grintena PD       0:00      1 (Priority)
```

Tested with commit 8d3bf03acbfb1a2d2561b8d22ee920448ff7bafa.

duylethanh:
This PR fixes #48.

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -70,6 +70,11 @@ class Config:
 		self._variants = OrderedDict()
 		self._exp_infos = OrderedDict()
 
+		self._experiments_discovered = False
+		self._experiments = OrderedDict()
+		self._runs_discovered = False
+		self._runs = OrderedDict()
+
 		def check_for_reserved_name(name):
 			if name.startswith('_'):
 				raise RuntimeError(f"Names starting with an underscore are reserved for internal simexpal objects: {name}")
@@ -202,9 +207,9 @@ class Config:
 			raise RuntimeError("Experiment {} does not exist".format(name))
 		return self._exp_infos[name]
 
-	def all_experiments(self):
-		def extract(selection):
-			# Helper to find all selected revisions for a given experiment.
+	def _fill_experiments(self):
+
+		def extract_experiments(selection):
 			def revisions_for_experiment(exp_info):
 				if 'use_builds' in exp_info._exp_yml:
 					if selection.revisions is not None:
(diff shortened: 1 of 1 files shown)


simexpal/base.py
```diff
@@ -218,10 +223,20 @@ def revisions_for_experiment(exp_info):
...
 
-		key = lambda x: (x[0].name, x[1].name if x[1] is not None else '_none',
-						 [sub_var.name for sub_var in x[2]])
-		for experiment_info, revision, variation in self._expand_matrix(extract, key=key):
-			yield Experiment(self, experiment_info, revision, variation)
+		if self._runs_discovered:
+			yield from self._runs.keys()
+
+		elif self._experiments_discovered:
+			yield from self._experiments.values()
```
avdgrinten, line 230: This is strange. I would expect that `all_experiments` only deals with experiments, not runs. Refactor this so that:
* If `not self._experiments_discovered`, some internal helper function is called to discover the experiments.
* Use the same helper function in `discover_all_runs()`. Note that that there can be experiments without runs but not the other way around. Hence, once the `_experiments` dict is built, `discover_all_runs()` can simply get its experiments from that dict (inserting into it is never necessary).

simexpal/base.py
```diff
@@ -247,12 +262,27 @@ def revisions_for_experiment(exp_info):
...
+									yield (self._experiments[(exp_info, revision, variation)], instance, rep)
+								else:
+									self._experiments[(exp_info, revision, variation)] = Experiment(self, exp_info, revision, variation)
+									yield (self._experiments[(exp_info, revision, variation)], instance, rep)
+			self._experiments_discovered = True
+
+		if self._runs_discovered:
+			for experiment in self._runs:
+				for instance, rep in self._runs[experiment]:
+					yield Run(self, experiment, instance, rep)
```
avdgrinten, line 275: Make `_runs` a single-level dict.

simexpal/base.py
avdgrinten: > There could be a case where discover_all_runs() was already called.

True, but then `_experiments` should be filled already. `discover_all_runs()` and `all_experiments()` should both call a (single) "fill `_experiments`" helper function.

simexpal/base.py
avdgrinten: No, I mean a `dict()` that just has `Run` objects as values (and tuples of Experiment/Instance/Repetition as keys, I guess).

simexpal/base.py

simexpal/base.py

PR #54, opened 2020-06-01
Support addition of extra paths to PATH environment variable/fix Issue #41

### LINKED ISSUE #41: Extra_paths variable to add additional paths to the PATH environment variable
We should possibly add an `extra_paths` variable for the experiments entries in the `experiments.yml` to add further paths to the PATH environment variable as noted in https://github.com/hu-macsy/simexpal/pull/36

duylethanh:
This PR fixes #41. This makes it possible to specify extra paths for experiments using the  `extra_paths` key, e.g.:
<pre>
experiments:
     - name: test
       args: ['my', 'experiment', 'arguments']
       extra_paths: ['/first/path', '/second/path']
<pre/>

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -666,6 +666,10 @@ class BuildInfo:
 	def regenerate(self):
 		return self._build_yml.get('regenerate', [])
 
+	@property
+	def extra_paths(self):
+		return self._build_yml.get('extra_paths', [])
+
 class Revision:
 	def __init__(self, cfg, revision_yml):
 		self._cfg = cfg
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -121,6 +121,10 @@ class RunManifest:
 	def workdir(self):
 		return self.yml['workdir']
 
+	@property
+	def builds(self):
+		return self.yml['builds']
+
 	def aux_file_path(self, ext):
 		return os.path.join(self.aux_subdir,
 				base.get_aux_file_name(ext, self.instance, self.repetition))
@@ -207,7 +211,8 @@ def compile_manifest(run):
 			'source': build.source_dir if build.revision.is_dev_build else build.clone_dir,
 			'compile': build.compile_dir,
 			'prefix': build.prefix_dir,
-			'exports_python': build.info.exports_python
+			'exports_python': build.info.exports_python,
+			'extra_paths': util.ensure_list_type(build.info.extra_paths)
 		}
 
 	# Collect extra arguments from variants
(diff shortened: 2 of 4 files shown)


avdgrinten:
Why is this bound to experiments and not to builds?

duylethanh:
> Why is this bound to experiments and not to builds?

There was a case where a user did not have their executable in `@THIS_PREFIX_DIR@/bin`,  so it might be useful to allow arbitrary paths in experiments. I will extend the PR to support the `extra_paths` variable for builds.

duylethanh:
It is now possible to specifiy `extra_paths` in the `regenerate`, `configure`,  `compile` and `install` key of `builds`, e.g.
<pre>
builds:
  - name: simexpal
    git: 'https://github.com/hu-macsy/simexpal'
    regenerate:
      - args: ['my', 'args']
        extra_paths: ['/first/path', '/second/path']
    configure:
      - args: ['my', 'args']
        extra_paths: ['/first/path', '/second/path']
    compile:
      - args: ['my', 'args']
        extra_paths: ['/first/path', '/second/path']
    install:
      - args: ['my', 'args']
        extra_paths: ['/first/path', '/second/path']
<pre/>

avdgrinten:
I would have expected the following: each build has `extra_paths` and these `extra_paths` apply to all experiments that use the build.

duylethanh:
> I would have expected the following: each build has `extra_paths` and these `extra_paths` apply to all experiments that use the build.

I've been trying to implement this and came across the following problem:
During the build process we allow the @-variables `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@`. This works perfectly fine for builds, but when running an experiment and invoking the run, we currently have no way of resolving the specific @-variables `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@` (they are actually not handled currently, as we have the @-variables  `@{PREFIX/COMPILE/SOURCE}_DIR_FOR:<build>@`).


My suggestions would be:
1. We deprecate the `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@` variables in favor of the existing `@{PREFIX/COMPILE/SOURCE}_DIR_FOR:<build>@` in the build process as they already include the `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@`-cases

or 

2. We add a `used_builds` property to the `RunManifest`-object and resolve `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@` to the respective directory of the build `RunManifest.used_builds[0]`

Do you agree with a suggestion or do you have something else?

avdgrinten:
Why not the following: we pass the used builds and their extra_paths through the manifest and merge them into PATH only inside invoke_run (while also resolving `@THIS_{PREFIX/COMPILE/SOURCE}_DIR@` at that point).

Basically like your second strategy but without restricting these variables to `used_builds[0]`.

duylethanh:
This PR is ready to be merged.

It is now possible to specify an `extra_paths` key for builds, e.g

```
builds:
  - name: simexpal
    git: 'https://github.com/hu-macsy/simexpal'
    configure: ...
    compile: ...
    install: ...
    extra_paths: ['/first/path', '/second/path']
   
```

Whenever an experiment (that uses this build) is started, simexpal will also check the paths specified in the `extra_paths` for the executable.

I also fixed the error message for unknown @-variables. Previously it would print out the whole string sequence.

duylethanh:
I created an extra `substitute_extra_paths` function to differentiate between substituting build `extra_paths` and experiments `args`, as they do not fully share the same @-variables.

PR is ready to be reviewed/merged.

PR #60, opened 2020-07-14
Impl. instance installation from git repo

AndreChinazzo:
Working in my machine.

In `experiments.yml`

```
...
instances:
  - gitrepo:
      url: <my repo url>
      commit_hash: <commit hash where to find the dataset file>
      filename: <dataset file name as in git repo>
    items:
      - 'myInstance'
...
```

Limitations: only one `items` in the list.

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -542,7 +542,14 @@ class Instance:
 				return
 
 		partial_path = os.path.join(self._cfg.instance_dir(), self.unique_filename)
-		if 'repo' in self._inst_yml:
+		if 'gitrepo' in self._inst_yml:
+			assert 'url' in self._inst_yml['gitrepo']
+			print("Downloading instance '{}' from {} repository".format(self.unique_filename,
+					self._inst_yml['gitrepo']['url']))
+			
+			instances.download_instance_git(self._inst_yml,
+					self.config.instance_dir(), self.unique_filename, partial_path, '.post0')
+		elif 'repo' in self._inst_yml:
 			print("Downloading instance '{}' from {} repository".format(self.unique_filename,
 					self._inst_yml['repo']))
 
--- a/simexpal/instances.py
+++ b/simexpal/instances.py
@@ -6,8 +6,10 @@ import shutil
 import tarfile
 import yaml
 import zipfile
+import subprocess
 
 from .util import try_mkdir
+from .util import try_rmtree
 
 class DownloadException(Exception):
 	pass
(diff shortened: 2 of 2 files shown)


duylethanh:
Thank you for the PR! I started looking into your request but you were faster than me.
If I see correctly the _**whole**_ git repository will be pulled, the desired file will be moved to the instance dir and the rest of the repo will be deleted afterwards. I am not sure if we should do it like that, especially since you mentioned Git LFS in your request. I thought about implementing something more general (not bounded to Git), which allows to download raw files from an URL. GitHub and Bitbucket offer raw views of files. For example: [this file](https://github.com/hu-macsy/simexpal/blob/master/scripts/simex) will be shown [like this](https://raw.githubusercontent.com/hu-macsy/simexpal/master/scripts/simex). In this way we can download only desired files and not the whole repo.

The `experiments.yml` could then look like this
<pre>
instances:
  - repo: raw
    url: 'url to the parent directory'
    items:
      - file1      # file name in the root directory
      - file2
</pre>

For example

<pre>
instances:
  - repo: raw
    url: 'https://raw.githubusercontent.com/hu-macsy/simexpal/master/simexpal/'
    items:
      - base.py
      - util.py
</pre>

would create the instances `base` and `util`, which are downloaded from `https://raw.githubusercontent.com/hu-macsy/simexpal/master/simexpal/base.py` and `https://raw.githubusercontent.com/hu-macsy/simexpal/master/simexpal/util.py` respectively. A current limitation to this approach is that items, which only differ in the extensions e.g. `foo.py`, `foo.cc`, can not be downloaded as simexpal will raise an exception due to the fact that they resolve to the same `shortname`.

What do you think @avdgrinten?

AndreChinazzo:
That would also works for me! I just need to check where to find the raw content in private gitlab repos.

BTW, I replied only to Alex, by mistake. But here follows it again:

The way I found to organize the datasets using LFS is to have multiple branches, where each branch has one dataset.
For example, I have one repo for the Netflix Prize dataset and subsets of it. The master branch contains only documentation, one branch has the whole dataset, and each of the other branches contains one subset of the whole thing. If I do some modification in any dataset, I simply make a new commit in the respective branch.

By using LFS, git will only download the whole content of LFS tracked files when you checkout to the intended commit. Otherwise, it download only a "pointer file".
As long as your LFS remote hard drive is big enough, you don't need to care about much else.

So, I thought about having an automatic clone and checkout to the intended commit (branch, version, ...), select the correct dataset file and clean up the rest.
This is probably a combination of what you already do in the building process (cloning, checking out) and the remote instance download (selecting dataset file, clean up).
Maybe so hash (eg. md5) would also be nice, to check whether an already existing file is indeed the correct one - but first simply downloading it again such do the trick.

avdgrinten:
I haven't looked deeply into this (will do later!). Note that sparse checkout could help to solve the issue Duy mentioned.

avdgrinten:
Thanks again for the PR. Duy and me thought about this for some more time; this functionality is now merged in #95, with slightly different syntax.

PR #65, opened 2020-08-14
Fixes #63

### LINKED ISSUE #63: simexpal should send SIGXCPU, then SIGKILL on timeout
Right now, we only send SIGXCPU. However, a process can catch SIGXCPU and continue running. After some grace period (maybe 30s by default?), we should kill the process with SIGKILL to ensure that the time limit is respected.

duylethanh:


simexpal/launch/common.py
```diff
@@ -374,7 +374,12 @@ def close(self):
 
 		elapsed = time.perf_counter() - start
 		if manifest.timeout is not None and elapsed > manifest.timeout:
-			child.send_signal(signal.SIGXCPU)
+			# programs can catch SIGXCPU and keep running
+			# kill command ensures that the time limit is respected (with a grace period)
```
avdgrinten, line 378: Coding style nitpick: comments should be proper sentences (capitalized and end with a dot).

duylethanh:
I made the requested changes. I forgot a comma after my initial fix so that I had to push another time. PR should be ready to be merged.

avdgrinten:
LGTM, thank you.

PR #68, opened 2020-09-01
Add documentation for extra_paths/experiments/variants/matrix key, at-variables and fix issues in Quick Start guide

duylethanh:
This PR contains documentation for the following things:

- ``extra_paths`` key in ``builds`` stanza
- ``experiments`` key
- ``variants`` key
- ``matrix`` key
- reference for @-variables

It also contains some fixes for issues in the Quick Start guide.

### DIFF
--- a/docs/at_variables_reference.rst
+++ b/docs/at_variables_reference.rst
@@ -1,0 +1,115 @@
+.. _AtVariables:
+
+@-Variables
+===========
+
+@-variables are placeholder variables that can be used in the ``experiments.yml`` file. They are enclosed
+by at signs, i.e. ``@...@`` and get resolved during runtime.
+
+For example: In order to avoid the duplication of command line arguments of experiments for each instance,
+e.g. ``experiment.py /path/to/instance1``, ``experiment.py /path/to/instance2``, ... we use the
+``@INSTANCE@`` variable that resolves to the respective paths of the instances. Then, we only need to
+specify ``experiments.py @INSTANCE@`` as experiment arguments.
+
+..
+    TODO: Add section Instances for generators and INSTANCE_FILENAME variable
+
+Below, we list all @-variables and where they can be used.
+
+- ``@COMPILE_DIR_FOR:<build_name>@``: :ref:`compilation directory <BuildDirectories>` of ``<build_name>`` in the same revision
+- ``@EXTRA_ARGS@``: extra arguments of all variants of an experiment
+- ``@INSTANCE@``: path of a :ref:`local <LocalInstances>`/:ref:`remote <RemoteInstances>` instance, i.e. ``/instance_directory/<instance_name>``
+- ``@INSTANCE:<ext>@``: path of a :ref:`MultipleExtensions` instance with extension ``<ext>``, i.e. ``/instance_directory/<instance_name>.<ext>``
+- ``@INSTANCE:<idx>@``: path of an :ref:`ArbitraryInputFiles` instance with index ``<idx>`` in the ``files`` key, i.e. ``/instance_directory/files[<idx>]``
+- ``@OUTPUT@``: path to the output file of an experiment
+- ``@OUTPUT:<ext>@``: path to the output file with extension ``<ext>`` of an experiment
+- ``@OUTPUT_SUBDIR@``: output subdirectory of the experiment where the output and status files are stored, i.e. ``/path_to_experiments_yml/output/``
+- ``@PARALLELISM@``: number of available CPUs
+- ``@PREFIX_DIR_FOR:<build_name>@``: :ref:`installation directory <BuildDirectories>` of ``<build_name>`` in the same revision
+- ``@REPETITION@``: repetition number of this experiment
+- ``@SOURCE_DIR_FOR:<build_name>@``: :ref:`source directory <BuildDirectories>` of ``<build_name>`` in the same revision
+- ``@THIS_CLONE_DIR@``: Deprecated for ``@THIS_SOURCE_DIR@``
+- ``@THIS_COMPILE_DIR@``: :ref:`compilation directory <BuildDirectories>` of this build
+- ``@THIS_PREFIX_DIR@``: :ref:`installation directory <BuildDirectories>` of this build
+- ``@THIS_SOURCE_DIR@``: :ref:`source directory <BuildDirectories>` of this build
+
+
+
+Builds
+------
+
+.. _AtVariablesBuildsCompile:
+
+compile
+^^^^^^^
+
+The following @-variables can be used in the ``compile`` key:
+
+- ``@COMPILE_DIR_FOR:<build_name>@``
+- ``@PARALLELISM@``
+- ``@PREFIX_DIR_FOR:<build_name>@``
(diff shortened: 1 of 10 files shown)


docs/at_variables_reference.rst
```diff
@@ -0,0 +1,93 @@
...
+
+@-Variables
+===========
+
+@-variables are placeholder variables that can be used in the ``experiments.yml`` file. They are enclosed
+by at signs, i.e. ``@...@`` and get resolved during runtime. Some use cases are:
+
+- allowing variations of experiments
+- accessing directories, which are created by simexpal during runtime
+- accessing values for keys specified in the ``experiments.yml``
```
avdgrinten, line 11: This is a bit unspecific. Can we maybe start with an example? One obvious use case is that you need @ variables to fill in the instance name.

docs/at_variables_reference.rst
```diff
@@ -0,0 +1,93 @@
...
+
+..
+    TODO: Add section Instances for generators and INSTANCE_FILENAME variable
+
+In the sections below we will list all @-variables and where they can be used.
```
avdgrinten, line 16: Maybe it makes sense to list all @ variables first and then list for each variable where it will be used.

docs/builds.rst
```diff
@@ -3,6 +3,11 @@
 Builds
 ======
 
+You might want to take a look at the following pages before continue reading:
```
avdgrinten, line 6: "... continue reading" -> exploring about automated builds.

docs/builds.rst
```diff
@@ -186,6 +191,29 @@ key to the respective dictionaries of the build steps.
 Specifying the working directory for other steps works analogously to specifying the working directory for the
 configuration step (as seen above).
 
+Extra Paths
+-----------
+
+If you do not want to install your executable in ``@THIS_PREFIX_DIR@/bin`` you need to specify the
```
avdgrinten, line 197: Add some rationale why you would want/wouldn't want to do this. `<prefix>/bin` is standard for many UNIX packages, but not all, especially not all academic software that sometimes uses custom build systems.

docs/experiments.rst
```diff
@@ -0,0 +1,318 @@
...
+Slurm
+-----
+
+.. _ExperimentsSupportedSlurmArgs:
+
+sbatch: ``--ntasks-per-node``, ``-c``, ``-N``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using a job scheduler like `Slurm <https://slurm.schedmd.com/overview.html>`_ it might be
+useful to benchmark your algorithms using different node/cpu settings.
```
avdgrinten, line 263: simexpal is not only for benchmarking algorithms but for running empirical software experiments in general.

docs/at_variables_reference.rst
duylethanh: Isn't it better to list for all keys the variables that can be used (as it is right now)? And not the other way around? I think that makes it easier to find the options you have for a key.

I think it is a good idea to list all @-variables beforehand though.

docs/builds.rst
duylethanh: Changed it for the other pages too (experiments, revisions, ...)

docs/at_variables_reference.rst
duylethanh: I removed those lines and added the ``@INSTANCE@``example

docs/builds.rst
duylethanh: done

docs/experiments.rst
duylethanh: changed "benchmark your algorithms" to "run your software"

duylethanh:
PR updated

docs/builds.rst
```diff
@@ -186,6 +191,31 @@ key to the respective dictionaries of the build steps.
 Specifying the working directory for other steps works analogously to specifying the working directory for the
 configuration step (as seen above).
 
+Extra Paths
+-----------
+
+For many UNIX packages it is standard to install the executable in the ``@THIS_PREFIX_DIR@/bin`` directory.
+This is why simexpal only checks those directories by default when looking for an executable. This might
+not always be the case, especially when using a custom build system. To cover those cases, we specify the
```
avdgrinten, line 199: Avoid two consecutive sentences starting with `this`.

avdgrinten: `... However, this assumption might not always be correct, for example, when using a custom build system.`

docs/experiments.rst
```diff
@@ -0,0 +1,318 @@
...
+
+- :ref:`QuickStart`
+- :ref:`AtVariables`
+- :ref:`Instances`
+- :ref:`Builds`
+- :ref:`Revisions`
+- :ref:`Variants`
+
+On this page we describe how to specify experiments in our ``experiments.yml`` file. We will see how
+to use builds in experiments and enable options such as repeating experiments or setting a timer.
```
avdgrinten, line 16: `[...] setting a timeout.`

docs/experiments.rst
```diff
@@ -0,0 +1,318 @@
...
+Moreover, we will see how to modify the ``sbatch`` command when using simexpal with Slurm and how to
+set environment variables.
```
avdgrinten, line 18: State this in a second paragraph. Slurm needs more introduction (that clears up that simexpal can be used together with batch scheduler -- not only Slurm). Not everybody will know what Slurm is.

docs/experiments.rst
```diff
@@ -0,0 +1,318 @@
...
+To specify an experiment we use the keys
+
+- ``name``: name of the experiment
+- ``args``: list of experiment arguments
+- ``output``: where experiment outputs should be stored.
+
+For now, choose ``stdout`` as value for the ``output`` key. Simexpal will then store the outputs
+of the experiments in the following directory:
+
+- ``/experiments.yml_directory/output/<experiment_name>~<variant_names>@<revision_name>``
```
avdgrinten, line 32: `experiments.yml_directory` looks a bit awkward. Can we just state relative paths?

avdgrinten (commented):
Some more comments; let's merge after these are addressed.

docs/builds.rst
duylethanh: done

docs/experiments.rst
duylethanh: Done. I also changed it in the "Timeout" section, as I used "timer" instead of "timeout" in some parts.

docs/experiments.rst
duylethanh: I stated that simexpal can be used with batch schedulers and stated Slurm and SGE as examples. (Although SGE is not really mentioned in the documentation at this point. I will add it later at some point.)

docs/experiments.rst
duylethanh: Done. As PR #69 was merged, I created a new section for the ``output``/``stdout`` key (the section can be found below the "Use Builds" section) and added the ``@OUTPUT:<ext>@`` to the at-variables.

duylethanh:
PR ready to be reviewed.

avdgrinten:
LGTM.

PR #69, opened 2020-09-17
Enhancement for 'output' key and addition of 'stdout' key in experiments stanza

duylethanh:
- It is possible to specify output extensions by adding the ``extensions`` key to the ``output`` dictionary, e.g. 
<pre>
output:
    extensions: [ext1, ext2]
</pre>
(``'out'`` will be added automatically)

- It is now possible to specify alternative output path extensions of experiments via the new ``stdout`` key by specifying the extension for the output file (the extensions has to be in the ``extensions`` list of ``output``), e.g.
```
experiments:
  - name: insertion-sort
    ...
    stdout: 'sol'
    output:
      extensions: ['sol']
```

 Then the results of the experiments will be written to 
``/path_to_exp_yml/output/<exp_name>~<variants>@<revision>/<inst_shortname>.sol``
- The new @-variable ``@OUTPUT:<ext>@`` is supported in the ``args`` key of experiments. This variable will resolve to ``/path_to_exp_yml/output/<exp_name>~<variants>@<revision>/<inst_shortname>.<ext>``
- Specifying ``'stdout'`` as value for the ``output`` key will trigger a DeprecationWarning
- ``simex e purge`` will now purge all respective output files (before, it would just purge the ``.out`` files)
- ``simex e print`` will now check the right output files for results (before, it would just check the ``.out`` files)
- ``Run.open_output_file()`` will now open the correct output file (before, it would just open the ``.out`` file)

### DIFF
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -6,6 +6,7 @@ import subprocess
 import time
 
 import yaml
+import warnings
 
 from .. import base
 from .. import util
@@ -98,10 +99,6 @@ class RunManifest:
 		return env_vars
 
 	@property
-	def output(self):
-		return self.yml['output']
-
-	@property
 	def timeout(self):
 		return self.yml['timeout']
 
@@ -124,6 +121,18 @@ class RunManifest:
 	@property
 	def builds(self):
 		return self.yml['builds']
+
+	@property
+	def output_extensions(self):
+		return self.yml['output_extensions']
+
+	@property
+	def stdout(self):
+		return self.yml['stdout']
+
+	@property
+	def variants(self):
+		return [v['name'] for v in self.yml['variants']]
 
 	def aux_file_path(self, ext):
 		return os.path.join(self.aux_subdir,
@@ -232,6 +241,12 @@ def compile_manifest(run):
 	for (k, v) in exp.info.environ.items():
 		environ[k] = str(v)
 
+	stdout = None
+	if exp.info.stdout is not None:
+		stdout = exp.info.stdout
+	elif exp.info.output == 'stdout':
+		stdout = 'out'
+
 	return RunManifest({
 		'config': {
 			'base_dir': run.config.basedir,
@@ -249,21 +264,38 @@ def compile_manifest(run):
 		'args': exp.info.args,
 		'timeout': float(exp.info.timeout) if exp.info.timeout is not None else exp.info.timeout,
 		'environ': environ,
-		'output': exp.info.output,
+		'output_extensions': exp.info.output_extensions,
+		'stdout': stdout,
 		'workdir': exp.info.workdir
 	})
 
 def invoke_run(manifest):
-	# Create the output file. This signals that the run has been started.
+
+	def get_qualified_output_file(ext):
+		if ext not in manifest.output_extensions:
+			raise RuntimeError(
+				f"Unexpected output extension for experiment '{manifest.experiment}': .{ext}\n"
+			)
+		return manifest.output_file_path(ext)
+
 	(stdout_pipe, stdout) = (None, None)
-	with open(manifest.output_file_path('out'), "w") as f:
+	if manifest.stdout is not None:
+		stdout_path = get_qualified_output_file(manifest.stdout)
+
 		# We do not actually need to write anything to the output file.
 		# However, we might want to pipe experimental output to it.
-		if manifest.output == 'stdout':
+		with open(stdout_path, 'w') as f:
 			stdout = os.dup(f.fileno())
-		else:
-			(stdout_pipe, stdout) = os.pipe()
-			os.set_blocking(stdout_pipe, False)
+	else:
+		stdout_path = manifest.aux_file_path('stdout')
+		(stdout_pipe, stdout) = os.pipe()
+		os.set_blocking(stdout_pipe, False)
+
+	# Create all the other the output files.
+	# The creation of the output file with extension '.out' signals that the run has been started.
+	diff_set = {manifest.stdout} if manifest.stdout is not None else set()
+	for ext in manifest.output_extensions - diff_set:
+		util.touch(manifest.output_file_path(ext))
 
 	# Create the error file.
 	(stderr_pipe, stderr) = os.pipe()
@@ -298,7 +330,9 @@ def invoke_run(manifest):
 		elif p == 'REPETITION':
 			return str(manifest.repetition)
 		elif p == 'OUTPUT':
-			return manifest.output_file_path('out')
+			return stdout_path
+		elif p.startswith('OUTPUT:'):
+			return get_qualified_output_file(p.split(':')[1])
 		elif p.startswith('SOURCE_DIR_FOR:'):
 			return manifest.get_source_dir_for(p.split(':')[1])
 		elif p.startswith('COMPILE_DIR_FOR:'):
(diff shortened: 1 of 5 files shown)


duylethanh:
PR ready to be reviewed.

simexpal/launch/common.py
```diff
@@ -6,10 +6,13 @@
 import time
 
 import yaml
+import warnings
 
 from .. import base
 from .. import util
 
+warnings.simplefilter(action='default', category=DeprecationWarning)  # do not ignore DeprecationWarnings
```
avdgrinten, line 14: When would the warning be ignored otherwise? IMHO, simexpal should not mess with global warning filters (because it can also be used as a library).

simexpal/launch/common.py
```diff
@@ -250,16 +265,61 @@ def compile_manifest(run):
...
+		return False
+
+	if manifest.output == 'stdout':
+		output_path = manifest.output_file_path('out')
+
+		msg = "Specifying the output path in the 'output' key is deprecated and will be removed in future " \
+				"versions. Use the 'stdout' key instead."
+		warnings.warn(msg, DeprecationWarning)
+	elif manifest.stdout is not None:
+		if manifest.stdout.startswith('@') and manifest.stdout.endswith('@'):
```
avdgrinten, line 303: This check looks strange. Why don't we expand @ variables unconditionally?

simexpal/launch/common.py
```diff
@@ -250,16 +265,61 @@ def compile_manifest(run):
...
+	else:
+		output_path = manifest.aux_file_path('stdout')
+
 	# Create the output file. This signals that the run has been started.
 	(stdout_pipe, stdout) = (None, None)
 	with open(manifest.output_file_path('out'), "w") as f:
 		# We do not actually need to write anything to the output file.
 		# However, we might want to pipe experimental output to it.
-		if manifest.output == 'stdout':
+		if output_path == manifest.output_file_path('out'):
```
avdgrinten, line 322: Why is this the right check? We always want to create the `.out` file eagerly (because status checks rely on it, right?). We might also want to create the other output files (= those with other extensions eagerly). The only case where we want to create a file lazily is the stdout file in aux. This file is only used if the user does *not* care about stdout (i.e., `stdout:` is not given).

simexpal/launch/common.py
duylethanh: The ``warnings`` module ignores ``DeprecationWarnings`` by default. Thats why I set ``action='default'``, which prints the first occurrence of the warning in each module. I actually set this warnings filter in ``simexpal/base.py`` already, so I might delete this line in ``simexpal/launch/common.py``

simexpal/launch/common.py
duylethanh: The ``stdout`` key takes 2 kinds of values.

- an @-variable

or

- some path

We could possibly remove the second case, as I'm not sure that it is practical since it expects a valid path for the exp/variants/rev/instance combination. This means, it will only be valid if the experiment only has one run. For all other runs the path will be invalid. Afterwards, we could expand the @-variable unconditionally like you suggested.

avdgrinten: Isn't the second case subsumed by just specifying `@OUTPUT@`?

avdgrinten: Actually, it might be better to just specify the extension of the output file. So `stdout: out` or similar.

simexpal/launch/common.py
avdgrinten: Hm, my point still stands: importing a library should not change global warning settings. We can set the filter in the `simex` script since in that script, we control the full environment.

avdgrinten: And it should also be removed from `launch/common.py`.

simexpal/launch/common.py
duylethanh: This check should be equivalent to the check that we had before. When ``manifest.output == 'stdout'``, then the output path is ``manifest.output_file_path('out'``). But now we have can achieve this via 2 different keys, namely the old ``'output'`` (with value ``'stdout'``)  and new ``'stdout'`` (with value ``@OUTPUT@`` or ``@OUTPUT:out@``)  key. Thats why I used the new variable ``output_path`` to catch both cases. The ``.out`` file will always we created, no matter the case (due to the ``with open(manifest.output_file_path('out'), 'w') as f:`` line that comes before the marked line)

simexpal/launch/common.py
duylethanh: Oh, I agree. I somehow overlooked the possibility to set the filter in the ``simex`` script. I will remove the line in the other files.

simexpal/launch/common.py
duylethanh: > Isn't the second case subsumed by just specifying `@OUTPUT@`?

Yes. I will remove this option.

> Actually, it might be better to just specify the extension of the output file. So `stdout: out` or similar.

Ok, sounds good. I will do the change

simexpal/launch/common.py
```diff
@@ -250,20 +263,47 @@ def compile_manifest(run):
...
+		if ext not in set(manifest.output_extensions + ['out']):
+			raise RuntimeError(
+				f"Unexpected output extension for experiment '{manifest.experiment}': .{ext}\n"
+			)
+		return manifest.output_file_path(ext)
+
+	is_directed = False
+
+	if manifest.output == 'stdout':
+		output_path = manifest.output_file_path('out')
```
avdgrinten, line 283: Rename `output_path` to `stdout_path` to clarify its purpose.

simexpal/launch/common.py
```diff
@@ -250,20 +263,47 @@ def compile_manifest(run):
...
-	# Create the output file. This signals that the run has been started.
+
+	def get_qualified_output_file(ext):
+		if ext not in set(manifest.output_extensions + ['out']):
+			raise RuntimeError(
+				f"Unexpected output extension for experiment '{manifest.experiment}': .{ext}\n"
+			)
+		return manifest.output_file_path(ext)
+
+	is_directed = False
```
avdgrinten, line 280: Rename to `redirect_stdout`.

avdgrinten: Or just set `output_path` to `None` to handle this case (is it ever used in the redirected case?).

simexpal/launch/common.py
```diff
@@ -250,20 +263,47 @@ def compile_manifest(run):
 		'timeout': float(exp.info.timeout) if exp.info.timeout is not None else exp.info.timeout,
 		'environ': environ,
 		'output': exp.info.output,
+		'output_extensions': exp.info.output_extensions,
+		'stdout': exp.info.stdout,
```
avdgrinten, line 267: Why do we need `output` and `stdout`? Isn't one of these keys enough?

simexpal/launch/common.py
duylethanh: done

simexpal/launch/common.py
duylethanh: I did the latter

duylethanh: The ``stdout_path`` is needed for the at-variable ``@OUTPUT@``. I was able to get rid of the extra boolean though.

simexpal/launch/common.py
duylethanh: I chose to keep the ``stdout`` key and removed the ``output`` key from ``RunManifest`` as it is not needed.

duylethanh:
PR ready to be reviewed.

duylethanh:
Fixed the merge conflicts. Now it is ready to be reviewed.

avdgrinten (approved):
LGTM

PR #70, opened 2020-09-25
Fix/issue 33

### LINKED ISSUE #33: Allow specification of machine environment (e.g., launcher) in experiments.yml
Specifying launchers on the command line can be error-prone. Add some functionality to determine the launcher from `experiments.yml`.

duylethanh:
- fix typo in error message and add more details
- added ``Experiment.repeat``property to avoid accessing protected class member
- added ``workdir`` key to ``experiments`` key in json schema
- added validation for the ``launchers.yml``
- resolves #33, i.e. it is possible to specify a ``launcher`` key that contains the name of a launcher specified in ``~/.simexpal/launchers.yml`` The ``launcher`` key can be specified in the respective entries of the ``experiments`` key

```
experiments:
  - name: quick-sort
    ...
    launcher: 'launcher1'
```

or  in the respective entries of the ``variants``key

```
variants:
  - axis: 'block-algo'
    items:
      - name: 'ba-insert'
        ...
        launcher: 'launcher2'
      - name: 'ba-bubble'
        ...
        launcher: 'launcher3'
  - axis: 'block-size'
    items:
      - name: 'bs32'
        ...
      - name: 'bs64'
        ...
```

If a launcher is specified in the ``experiments`` key and the ``variants`` key, the launcher specified in the variant will be used.
Specifying a launcher in 2 different axes will raise a ``RuntimeError`` if an experiment uses both variant axes.


avdgrinten:
Hm, I think this is not really what we want. This mixes the per-cluster/environment configuration from `launchers.yml` and the more-or-less portable definition of experiments in `experiments.yml`. What we want instead is a `launcher: foobar` key, where foobar is the name of a launcher that is defined in `launchers.yml`.

duylethanh:
Ready to be reviewed. I also added validation for the ``launchers.yml``.

avdgrinten:
Why do we need `append_to_launchers_dict`? Isn't there only ever one launcher in that dict? That function looks a bit inflexible to me, what if we have launchers that do not only depend on the `scheduler` and `queue` properties? Isn't that the case already for SLURM since it also depends on extra args for sbatch?

duylethanh:
PR updated and ready to be reviewed.
The keys of the ``launchers`` dict are the names of the launchers now. ``append_to_launchers_dict`` now takes the launcher name, the run to be appended and a variable number of args (``*args``) as input. ``*args`` is used for the ``create_launcher()`` call. In the future we might want to use ``**kwargs``, if we allow a variety of launchers that depend on many different parameters.

avdgrinten (approved):
LGTM.

PR #73, opened 2020-10-06
Resolve #72

### LINKED ISSUE #72: slurm_args for variation
It would be nice to support slurm_args also for variations.

Context: I needed to pass the '--exclusive' parameter to slurm only for a specific variant that used less that half of the cores of 
a node and without it, slurm assigned 2 jobs in one node.

duylethanh:


avdgrinten (approved):
LGTM. In the future, we might want to think about a more generic way to handle "effective" parameters.

PR #75, opened 2020-11-04
Add caching to status checks

duylethanh:
Simexpal will now create a ``status.cache`` file in every ``experiments.yml`` directory and cache the status of every run in it. The cache will be updated at the end of every ``simex e`` command (e.g. ``launch``, ``list``, ...).

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -684,6 +684,8 @@ def do_experiments_purge(args):
 				for ext in run.experiment.info.output_extensions:
 					util.try_rmfile(run.output_file_path(ext))
 
+				run.purge_status_cache_dict()
+
 			else:
 				print("This would purge experiment '{}', instance '{}' [{}]. Use '-f' to confirm this action.".format(
 						exp.name, instance, run.repetition))
--- a/scripts/simex
+++ b/scripts/simex
@@ -493,6 +493,8 @@ def do_experiments_list(args, as_default_subcmd=False):
 			show_compact_list(args.full)
 	print(len(selection), "experiments in total")
 
+	cfg.writeback_status_cache()
+
 experiments_list_parser = experiments_subcmds.add_parser('list',
 		parents=[run_selection_parser])
 experiments_list_parser.set_defaults(cmd=do_experiments_list)
@@ -538,6 +540,8 @@ def do_experiments_info(args):
 	print("All related variants are:")
 	for v in sorted(list(variants_name_set)):
 		print('\t', v)
+
+	cfg.writeback_status_cache()
 
 experiments_info_parser = experiments_subcmds.add_parser('info',
 		parents=[run_selection_parser])
@@ -643,6 +647,8 @@ def do_experiments_launch(args):
 
 		submit_to_launcher(cfg, launcher_runs)
 
+		cfg.writeback_status_cache()
+
 experiments_launch_parser = experiments_subcmds.add_parser('launch',
 		parents=[run_selection_parser])
 experiments_launch_parser.set_defaults(cmd=do_experiments_launch)
@@ -682,6 +688,8 @@ def do_experiments_purge(args):
 				print("This would purge experiment '{}', instance '{}' [{}]. Use '-f' to confirm this action.".format(
 						exp.name, instance, run.repetition))
 
+	cfg.writeback_status_cache()
+
 experiments_purge_parser = experiments_subcmds.add_parser('purge',
 		parents=[run_selection_parser])
 experiments_purge_parser.set_defaults(cmd=do_experiments_purge)
@@ -700,6 +708,8 @@ def do_experiments_print_output(args):
 			print('Instance: {}'.format(run.instance.shortname))
 			print('Output:\n\n{}\n'.format(util.read_file(run.output_file_path_from_yml())))
 			print('Error Output:\n\n{}\n'.format(util.read_file(run.aux_file_path('stderr'))))
+
+	cfg.writeback_status_cache()
 
 experiments_print_parser = experiments_subcmds.add_parser('print',
 		parents=[run_selection_parser])
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -2,9 +2,11 @@
 from collections import OrderedDict
 from enum import IntEnum
 import itertools
+import json
 import os
 import yaml
 import sys
+import tempfile
 import warnings
 
 from . import instances
@@ -76,6 +78,9 @@ class Config:
 		self.basedir = basedir
 		self.yml = yml
 
+		self.status_cache_path = basedir + '/status.cache'
+		self.status_cache_dict = {}
+
 		self._insts = OrderedDict()
 		self._build_infos = OrderedDict()
 		self._revisions = OrderedDict()
@@ -145,6 +150,12 @@ class Config:
 				if exp_yml['name'] in self._exp_infos:
 					raise RuntimeError("The experiment name '{}' is ambiguous".format(exp_yml['name']))
 				self._exp_infos[exp_yml['name']] = ExperimentInfo(self, exp_yml)
+
+		try:
+			with open(self.status_cache_path, 'r') as f:
+				self.status_cache_dict = json.load(f)
+		except FileNotFoundError:
+			pass
 
 	def instance_dir(self):
 		"""Path of the directory that stores all the instances."""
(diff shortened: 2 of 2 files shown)


simexpal/base.py
```diff
@@ -522,6 +533,55 @@ def make_variation(prod):
 
 		return [make_variation(prod) for prod in itertools.product(*variation_bundle)]
 
+	def update_status_cache_dict(self, run):
+
+		def _get_status_entry(run):
```
avdgrinten, line 538: Since this function is only called once, we can inline it into `update_status_cache_dict`.

simexpal/base.py
```diff
@@ -1129,25 +1189,33 @@ def output_file_path(self, ext):
...
-		elif os.access(self.output_file_path('out'), os.F_OK):
-			return Status.STARTED
-		elif os.access(self.aux_file_path('run'), os.F_OK):
-			return Status.SUBMITTED
-		elif os.access(self.aux_file_path('lock'), os.F_OK):
-			return Status.IN_SUBMISSION
-
-		return Status.NOT_SUBMITTED
+
+		def _get_last_mod():
```
avdgrinten, line 1193: Likewise, this function can be inlined into the caller. (These are micro optimizations, of course, but get_status() is also called a lot and function calls are not entirely free in Python, compared to other languages.)

simexpal/base.py
```diff
@@ -522,6 +533,55 @@ def make_variation(prod):
...
+				return Status.SUBMITTED, os.stat(run.aux_file_path('run')).st_mtime
+			elif os.access(run.aux_file_path('lock'), os.F_OK):
+				return Status.IN_SUBMISSION, os.stat(run.aux_file_path('lock')).st_mtime
+			return Status.NOT_SUBMITTED, None
+
+		status, last_mod = _get_status_entry(run)
+
+		if run.experiment.display_name not in self.status_cache_dict:
+			# Stringify key tuple as json can not serialize pure tuples.
+			self.status_cache_dict[run.experiment.display_name] = {
```
avdgrinten, line 565: We use `display_name` here. That is probably not the best choice; we could add another `internal_name` (or similar) function that also incorporates the repetition. That way, we would not need a nested dict.

avdgrinten: Note that we also form `display_name` multiple times, which might involve more string concatenations / memory allocations than necessary (since I doubt that the Python interpreter is able to optimize these calls).

scripts/simex
```diff
@@ -678,10 +684,14 @@ def do_experiments_purge(args):
 				for ext in run.experiment.info.output_extensions:
 					util.try_rmfile(run.output_file_path(ext))
 
+				cfg.purge_status_cache_dict(run)
```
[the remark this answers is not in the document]
duylethanh: Yes, this should not affect the correctness of the caching. This function is intended to clear the cache from entries that we do not need anymore i.e. the experiments, which we purge.

simexpal/base.py
```diff
@@ -14,6 +16,8 @@
 EXPERIMENTS_LIST_THRESHOLD = 30
 TIMEOUT_GRACE_PERIOD = 30
 
+STATUS_CACHE_PATH = './status.cache'
```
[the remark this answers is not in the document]
duylethanh: I will move the variable to the ``Config`` class in order to do so

simexpal/base.py
```diff
@@ -522,6 +533,55 @@ def make_variation(prod):
 
 		return [make_variation(prod) for prod in itertools.product(*variation_bundle)]
 
+	def update_status_cache_dict(self, run):
```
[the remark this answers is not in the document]
duylethanh: Currently this function is called from the ``Run`` class, so I think it should be public. Or should I still change it? Maybe we should move the function then.

simexpal/base.py
duylethanh: Ok, I will add an ``Experiment.internal_name`` property that I will use for the cache entry keys.

simexpal/base.py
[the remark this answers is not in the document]
duylethanh: I moved the function to the ``Run`` class.

simexpal/base.py

simexpal/base.py

simexpal/base.py
duylethanh: I added an ``internal_name`` property to the ``Run`` class. It works like ``Experiment.display_name``, but adds a suffix ``[<repetition>]``.

duylethanh:
PR updated. I implemented the change requests and further added  missing ``Config.writeback_status_cache()`` calls to functions that use ``Run.get_status()``(i.e. ``Config.collect_successful_results()`` and ``Config.export_experiments()``)

simexpal/base.py
```diff
@@ -1117,6 +1137,17 @@ def __init__(self, cfg, experiment, instance, repetition):
 	def config(self):
 		return self._cfg
 
+	@property
+	def internal_name(self):
+		internal_name = self.experiment.name
+		if self.experiment.variation:
+			internal_name += ' ~ ' + ', '.join([variant.name for variant in self.experiment.variation])
```
avdgrinten, line 1144: Drop the spaces in `internal_name` (also for the revision).

simexpal/base.py
```diff
@@ -1128,26 +1159,71 @@ def output_file_path(self, ext):
...
+				status, last_mod = Status.STARTED, os.stat(self.output_file_path('out')).st_mtime
+			elif os.access(self.aux_file_path('run'), os.F_OK):
+				status, last_mod = Status.SUBMITTED, os.stat(self.aux_file_path('run')).st_mtime
+			elif os.access(self.aux_file_path('lock'), os.F_OK):
+				status, last_mod = Status.IN_SUBMISSION, os.stat(self.aux_file_path('lock')).st_mtime
+
+		internal_name = self.internal_name
+		if internal_name not in self._cfg.status_cache_dict:
+			self._cfg.status_cache_dict[internal_name] = {
+				self.instance.shortname: {'status': status, 'last_mod': last_mod}}
```
avdgrinten, line 1190: Is `shortname` enough to uniquely identify an instance?

simexpal/base.py
```diff
@@ -1128,26 +1159,71 @@ def output_file_path(self, ext):
...
+					try:
+						last_mod = os.stat(self.aux_file_path('lock')).st_mtime
+					except FileNotFoundError:
+						pass
+
+		internal_name = self.internal_name
+		if internal_name in self._cfg.status_cache_dict:
+			if self.instance.shortname in self._cfg.status_cache_dict[internal_name]:
+				cache_entry = self._cfg.status_cache_dict[internal_name][self.instance.shortname]
+				if last_mod == cache_entry["last_mod"]:
```
avdgrinten, line 1224: Small optimization: we only need to compute `last_mod` here, i.e., before the `if`. This should make it somewhat faster to add new entries to the cache.

avdgrinten: This code could potentially break if the modification time of two files is exactly identical (e.g., `.status` was not found yet but has the same timestamp as `.lock`). That can probably not happen on modern file systems with sub-second precision but it might be worth to add a comment here; we do not need to fix it now.

simexpal/base.py
duylethanh: It should be enough. When constructing the instance dictionary ``Config._insts`` we check for instances that resolve to the same ``shortname`` (and raise an error in case it does)

simexpal/base.py
duylethanh: Done

simexpal/base.py
duylethanh: > Small optimization: we only need to compute `last_mod` here, i.e., before the `if`. This should make it somewhat faster to add new entries to the cache.

Done.



> This code could potentially break if the modification time of two files is exactly identical (e.g., `.status` was not found yet but has the same timestamp as `.lock`). That can probably not happen on modern file systems with sub-second precision but it might be worth to add a comment here; we do not need to fix it now.

To fix it, we could also save the file extension of the file that produced this timestamp. Although the case above should not really occur, except when the user manually deletes status related files **and** those files have the same timestamp.

duylethanh:
PR updated.

simexpal/base.py
avdgrinten: Alright.

simexpal/base.py
avdgrinten: True. We probably do not even need to store additional info, we can reconstruct from the cached status which file we read (and compare that to the last_mod file).

duylethanh: I have added another check so that the above scenario can not happen.

duylethanh:
PR updated. I added code to prevent the code from breaking in case files have the same timestamp and one file got deleted.

avdgrinten (approved):
LGTM now, thank you.

Issue #77, opened 2020-11-13
Overhaul queue launcher, remove evloop.py


avdgrinten:
The queue launcher is currently based on a custom event loop that is used only in a single place. It needs to be revamped. The code can probably be simplified by using Python's `selectors` module directly. To actually run the experiments, it is probably easier to launch `simex internal-invoke` in a different process.

- [ ] Remove evloop.py and use `selectors` directly instead (we already do that in `launch/common.py`).
- [ ] Instead of calling into `launch.common.invoke_run`, use a subprocess that executes `simex internal-invoke` on a manifest file.
- [ ] Integrate that subprocess into the event loop such that we can take additional requests from the socket while the process is running.
- [x] Add functionality to launch the queue daemon automatically. Details need to be discussed.

avdgrinten:
After think about it a bit more, the last bullet sounds quite tedious to implement due to various race conditions around `bind()`ing and `unlink()`ing UNIX sockets and launching the daemon itself. It might be better to provide a systemd unit file to start the daemon.

duylethanh:
Closing this issue, as the respective PR #80 has been merged with commit 95fe6c6c285c87f6396399d13ecd484385c21f55.

PR #80, opened 2020-11-23
Overhaul QueueLauncher (fix issue #77)

duylethanh:
The PR implements the points stated in issue #77:
- Remove evloop.py and use selectors directly instead (we already do that in launch/common.py).
- Instead of calling into ``launch.common.invoke_run()``, use a subprocess that executes ``simex internal-invoke`` on a manifest file.
- Integrate that subprocess into the event loop such that we can take additional requests from the socket while the process is running.


duylethanh:
I deleted the ``evloop.py`` and removed the import in ``scrips/simex``

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
-		req = util.yaml_from_string(self.recv_buffer.decode())
-		resp = self.queue.dispatch(descriptor.get_loop(), req)
-		self.state = _State.DONE
-		self._handle.unregister()
+	def __init__(self, serve_socket):
+		self.socket = serve_socket
+		self.socket_path = serve_socket.getsockname()
+		self.selector = selectors.DefaultSelector()
+		self.subprocess = None
+		self.subprocess_terminated = True
```
avdgrinten, line 19: Why is this not a local variable in the run() method?

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
+				if data:
+					self.requests.append(data)
+				sk.data.close()
+				self.selector.unregister(sk.fd)
+
+			try:
+				self.subprocess_terminated = self.subprocess.poll() is not None
+			# No subprocess has been started yet.
+			except AttributeError:
+				pass
```
avdgrinten, line 61: Why don't we just check `if self.subprocess is not None` instead of catching this exception?

simexpal/launch/queue.py
```diff
@@ -1,18 +1,33 @@
...
+			util.write_yaml_file(f, specs)
+
 		print("Launching experiment '{}', instance '{}' on local machine".format(
 				run.experiment.name, run.instance.shortname))
 
 		queuesock.sendrecv({
 			'action': 'launch',
-			'manifest': common.compile_manifest(run).yml
+			'manifest': manifest,
+			'specfile_path': specfile
```
avdgrinten, line 31: I dislike that we submit both the manifest and the specfile. We should not store information in a redundant way. We should probably just pass the file name and read the file from the daemon process.

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
-		if data:
-			self.recv_buffer += data
-			return
+class Queue:
 
-		req = util.yaml_from_string(self.recv_buffer.decode())
-		resp = self.queue.dispatch(descriptor.get_loop(), req)
-		self.state = _State.DONE
-		self._handle.unregister()
+	def __init__(self, serve_socket):
```
avdgrinten, line 14: Make sure that private variables start with an underscore here.

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
+					display_name = self.get_display_name(request)
+					print("Launching run {}/{}[{}]".format(
+						display_name, request['manifest']['instance'], request['manifest']['repetition']))
+
+					script = os.path.abspath(sys.argv[0])
+					specfile = request['specfile_path']
+
+					self.subprocess = subprocess.Popen([script, 'internal-invoke', '--queue', specfile])
+
+				elif request['action'] == 'stop':
```
avdgrinten, line 76: I think we should only push experiments to the request queue and not other actions. Stop actions can be processed by setting a `._should_stop` variable that we check at the appropriate time.

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
+					print("Closing socket on {}".format(self.socket_path))
+					self.socket.shutdown(socket.SHUT_RDWR)
+					self.socket.close()
```
avdgrinten, line 79: shutdown() before close() is not necessary.

avdgrinten (commented):
Below are a few comments but the major thing that needs to be changed is the handling of `accept()`.

* When run() starts, we should add the `Queue`'s `socket` to the selector.
* When that socket becomes readable, we should call `accept()`.
* We should pass the new socket (that `accept()` returned) to the `Connection` constructor. Then, we should also add the new socket to the selector and only call `.progress()` if the new socket is readable. Right now, we just performs reads in `.progress()` without checking if the socket is readable or not; this will block the entire process.

scripts/simex
```diff
@@ -829,6 +826,7 @@ def do_invoke(args, basedir=None):
 invoke_parser = main_subcmds.add_parser('internal-invoke')
 invoke_parser.set_defaults(cmd=do_invoke)
 invoke_parser.add_argument('-n', action='store_true')
+invoke_parser.add_argument('--queue', action='store_true')
 invoke_parser.add_argument('--slurm', action='store_true')
```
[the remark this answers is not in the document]
duylethanh: I added a ``--method`` argument, that takes a value in ``['queue', 'slurm, 'slurm-array', 'sge-index']``

simexpal/queuesock.py
duylethanh: I changed ``self._subprocess`` and ``self._subprocess_terminated`` to private variables and further added ``self._should_stop``

simexpal/queuesock.py
```diff
@@ -1,105 +1,121 @@
...
+				else:
+					print("Ignoring request with unknown action '{}': {}".format(request['action'], request))
+
+class Connection:
+	def __init__(self, socket):
+		self.socket = socket
+		self.connection = None
+
+	def progress(self):
+		self.connection = self.socket.accept()[0]
```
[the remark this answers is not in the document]
duylethanh: Done

simexpal/queuesock.py
duylethanh: Done

simexpal/queuesock.py
duylethanh: Done. (It came from an older approach)

simexpal/launch/queue.py
duylethanh: Done

simexpal/queuesock.py
duylethanh: I changed ``self._subprocess`` and ``self._subprocess_terminated`` to private variables and further added ``self._should_stop``

simexpal/queuesock.py
duylethanh: Done

simexpal/queuesock.py
duylethanh: Done

duylethanh:
I made the requested changes. I also added another commit to fix the broken ``internal-invoke`` command for the SgeLauncher (unfortunately, we currently have no SGE cluster to test this, but the command was definitely faulty)

avdgrinten (commented):
`_subprocess_terminated` is still a member variable and not a local variable. Otherwise, this LGTM.

duylethanh:
> 
> 
> `_subprocess_terminated` is still a member variable and not a local variable. Otherwise, this LGTM.

``cur_subprocess`` (formerly ``self.subprocess``) and ``subprocess_termianted`` are now local variables

avdgrinten (approved):
LGTM.

Issue #82, opened 2020-11-26
Allow variants without extra_args


avdgrinten:
Simexpal complains about variants without extra_args even though they are perfectly reasonable (e.g., if they set environment variables):

```
simexpal: Validation error in experiments.yml at [variants][0][items][0]:
{'environ': 'OMP_NUM_THREADS':1, 'name': 't01'}
'extra_args' is a required property
```

PR #83, opened 2020-11-26
Remove 'extra_args' from required keys in variants

duylethanh:
Fix issue #82.


avdgrinten:
So the code did not actually require `extra_args` to be present and this was only forced by the schema? If that is the case, this LGTM.

duylethanh:
> So the code did not actually require `extra_args` to be present and this was only forced by the schema? If that is the case, this LGTM.

Yes, this was forced by the schema

PR #84, opened 2020-11-27
fileless instances (issue #78) and instances with extra arguments

### LINKED ISSUE #78: Fileless Instances
We should add the possibility of fileless instances as an instance might be defined through input parameters ,e.g, algorithms that generate their data themselves and only need an input parameter like ``--seed=X``.

This might be done by adding an ``args`` key to instances and a new at-variable ``@INSTANCE_ARGS@`` that resolves to the respective instance args.

duylethanh:
- restructuring/fixing the ``schemes/experiments.json`` schema and addition of ``extra_args``to the ``instances`` stanza
- support of fileless instances( resolves #78), by introducing the ``extra_args`` key for the ``instances``stanza
- documentation for fieless instances and the ``extra_args`` key

Below are some more details for the first two points:

The ``extra_args`` of instances are accessible via the already existing ``@EXTRA_ARGS@`` variable, which only resolved to the ``extra_args`` of variants before. Now it resolves to the concatenation of both (possible more, because we can have more than one variant for an experiment)``extra_args`` (from the instance and the variants).

-----

The validation schema for instances in the ``experiments.yml`` file can now roughly be expressed as:

```
if 'generator' in instance:
	check syntax for generator instances
if 'extensions' in instance:
	check syntax of multiple extension instances
if 'files' in instance:
	check syntax of arbitrary file instances
if neither of the above cases:
	check syntax for local/remote
```

(Notice the ``if-if``-construct instead of an ``if-elif``-construct which is due to the fact that [jsonschema 7.0](https://json-schema.org/understanding-json-schema/reference/conditionals.html) does not currently support ``if-elif``constructs.)

----

If an entity fails a ``oneOf`` construct, simexpal will now print out the errors for each subschema, e.g,: 

the entity

```
instances:
  - repo: local
    items:
      - name: baz
        extra_args: ['some','extra_args']
      # 'name' key is missing and 'foo' is an invalid key
      - extra_args: ['some','extra_args']
        foo: bar
  - repo: locjal    # typo of local
    items:
      - instance1
```

produces the following output when calling a ``simex`` command:

Screen Shot 2020-11-27 at 21 57 50

### DIFF
--- a/docs/experiments.rst
+++ b/docs/experiments.rst
@@ -52,6 +52,12 @@ a list of strings (instead of one space separated string). Note that the :ref:`@
 ``@INSTANCE@`` resolves to the paths of the instances given in the ``instances`` stanza, i.e,
 ``/instance_directory/<instance_name>``.
 
+If the instances have :ref:`InstanceExtraArguments`, we further need to add the :ref:`@-variable <AtVariables>`
+``@EXTRA_ARGS@`` to the experiment arguments, e.g,
+``args: ['./sort.py', '--algo=insertion-sort', '@INSTANCE@', '@EXTRA_ARGS@']`` for the example above.
+``@EXTRA_ARGS@`` will resolve to the specified extra arguments of the respective instance (and also the used
+:ref:`variants <ExperimentsWithVariants>`, see below).
+
 .. _ExperimentsWithMultipleExtensionInstances:
 
 Experiments with Multiple Extension Instances
@@ -78,6 +84,8 @@ specify your experiment as follows:
 The ``@INSTANCE:graph@`` variable will resolve to ``/instance_directory/<instance_name>.graph`` during
 runtime. Analogously for the ``@INSTANCE:xyz@`` variable.
 
+:ref:`InstanceExtraArguments` are handled analogously to the case of :ref:`ExperimentsWithLocalRemoteInstances`.
+
 Experiments with Arbitrary Input File Instances
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -103,6 +111,10 @@ as follows:
 The ``@INSTANCE:0@`` variable will resolve to ``/instance_directory/files[0]``, where ``files[0]`` is
 the first filename of the ``files`` key. Analogously for the ``@INSTANCE:1`` variable.
 
+:ref:`InstanceExtraArguments` are handled analogously to the case of :ref:`ExperimentsWithLocalRemoteInstances`.
+
+.. _ExperimentsWithVariants:
+
 Experiments with Variants
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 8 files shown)


docs/experiments.rst
```diff
@@ -103,6 +115,14 @@ as follows:
 The ``@INSTANCE:0@`` variable will resolve to ``/instance_directory/files[0]``, where ``files[0]`` is
 the first filename of the ``files`` key. Analogously for the ``@INSTANCE:1`` variable.
 
+If the instances have :ref:`InstanceExtraArguments`, we further need to add the :ref:`@-variable <AtVariables>`
+``@EXTRA_ARGS@`` to the experiment arguments, e.g,
+``args: ['./algorithm.py', '@INSTANCE:0', '@INSTANCE:1@', '@EXTRA_ARGS@']`` for the example above.
+``@EXTRA_ARGS@`` will resolve to the specified extra arguments of the respective instance (and also the used
+:ref:`variants <ExperimentsWithVariants>`, see below).
```
avdgrinten, line 122: I don't link that this paragraph is repeated three times. Can we shorten the repetitions and refer back to the first occurrence?

simexpal/util.py
```diff
@@ -97,6 +97,19 @@ def _validate_dict(dictionary, source, schema_path):
 
 				print("simexpal: Validation error in {} at {}:\n{}\n{}".format(
 					source, err_source, err.instance, err.message), file=sys.stderr, end="\n\n")
+
+				# The error comes from subschemas in anyOf, oneOf or allOf.
+				if err.context:
```
avdgrinten, line 102: `context` is only set for anyOf, oneOf and allOf? Did you test that is does not give unreasonable output in other cases?

simexpal/launch/common.py
```diff
@@ -256,9 +262,11 @@ def compile_manifest(run):
 		'variants': variants_yml,
 		'revision': exp.revision.name if exp.revision else None,
 		'instance': run.instance.shortname,
-		'instance_filename': run.instance.yml_name,
+		'instance_yml_name': run.instance.yml_name,
```
avdgrinten, line 265: Why not just `instance_name`?

avdgrinten (commented):
Looks mostly good, a few minor comments below.

simexpal/util.py
duylethanh: > `context` is only set for anyOf, oneOf and allOf? 

Yes, that should be the case. The following text comes from the [official documentation](https://python-jsonschema.readthedocs.io/en/stable/errors/): _With validators like anyOf, the ``ValidationError.context`` attribute can be used to see the sub-errors which caused the failure._

> Did you test that is does not give unreasonable output in other cases?

Yes, I did some testing and in other cases the ``context`` attribute is just an empty list ``[]``.

simexpal/launch/common.py
duylethanh: I changed it to ``instance_name``

docs/experiments.rst
duylethanh: I removed the repetition and references the first occurence

duylethanh:
PR updated.

avdgrinten:
LGTM.

PR #86, opened 2020-11-30
Remove download option for konect instances

### LINKED ISSUE #85: Download of konect instances no longer possible.
As [konect.cc](http://konect.cc) is now only accessible via a paid subscription, we probably can not offer the possibility to download konect instances anymore. Trying to download konect instances currently causes simexpal to crash with a ``ConnectionError``.

duylethanh:
Resolves #85.

One thing that we should maybe reconsider is the removal of the konect case in ``convert_to_edgelist()`` located in the ``instances.py``. Users might have already downloaded konect instances. 

Alternatively, we could only remove the option of downloading konect instances and leave the rest. In this case users can still use ``simex instances run-transform`` manually on the instances.

### DIFF
--- a/docs/instances.rst
+++ b/docs/instances.rst
@@ -9,10 +9,9 @@ You might want to take a look at the following pages before exploring instances:
 
 On this page we describe how to specify instances in the ``experiments.yml`` file. You can
 list local instances that consist of one file or several files. More over simexpal can download
-remote instances from the `KONECT <http://konect.cc>`_ and `SNAP <https://snap.stanford.edu/data/>`_
-repository. It is also possible to assign instances to instance sets that enable a more efficient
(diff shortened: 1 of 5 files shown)


avdgrinten:
I think in the long term, we should move the graph-specific code into some kind of plugin system since Simexpal should not only focus on graphs but on arbitrary algorithmic experiments.

For now the second option (only disable downloading, keep the remaining logic) sounds reasonable.

duylethanh:
PR updated. I removed the download option for konect instances and adjusted the documentation accordingly. I also added the ``raise`` keyword to the ``DownloadException`` as it was just hanging around.

avdgrinten (approved):
LGTM.

PR #89, opened 2020-12-02
Add Github Actions workflow and unit test for 'util.validate_setup_file()' and cli

duylethanh:
- Github Actions workflow triggered by pull requests on the ``master`` branch
- unit tests (with ``pytest``) for ``util.validate_setup_file()``, ``simex e launch --launch-through=fork`` and ``simex e purge --all -f``
- ~add `` pytest6.1.2`` to ``install_requires`` in ``setup.py``~

The workflow currently uses the cross product of ``[ubuntu-latest, macos-latest]`` and ``[python3.6, python3.7]`` to run the unit tests The following screenshot is from a test pull request on my fork https://github.com/duylethanh/simexpal/pull/4:

GithubActions

Some things to consider:

- `` ubuntu-latest`` resolves to Ubuntu 18.04, although Ubuntu 20.04 is available as preview
- ``macos-latest`` resolves to macOS 10.15 (Catalina), although macOS 11.0 (BigSur) is available as preview
- currently the workflow takes roughly 30s-1min to run
- (for the future) in case the number of dependencies increases, we should consider caching
- I have stated a specific version of ``pytest``, we might wanna relax that condition


pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+minversion = 6.0
+testpaths=
+    tests
+
```
avdgrinten, line 5: Extra empty line.

setup.py
```diff
@@ -25,7 +25,8 @@
 		'argcomplete',
 		'requests',
 		'pyyaml',
-		'jsonschema>=3.2.0'
+		'jsonschema>=3.2.0',
+		'pytest==6.1.2'
```
avdgrinten, line 29: Do we need pytest as a dependency? This should not be necessary for end users, right?

pytest.ini
avdgrinten: Also in other files.

tests/validation/test_experiments_yml.py
```diff
@@ -0,0 +1,30 @@
+
```
avdgrinten, line 1: Whitespace is a bit strange in this file (empty line after function declaration, double blank lines?).

avdgrinten (commented):
Looks good so far. ubuntu-latest and macos-latest is fine, we do not necessarily need to test for bleeding edge images.

Are there any other simple but useful tests that we could add? Maybe running experiments via the Python API (= a fork launcher)? I guess we would have to copy the experiments directory to a temporary location for that to work.

pytest.ini
duylethanh: I removed the extra blank lines at the end of all respective files

setup.py
duylethanh: Indeed, ``pytest`` is not necessary for end users

tests/validation/test_experiments_yml.py
duylethanh: Fixed

duylethanh:
PR updated. I made the requested changes, while doing some refactoring to the new code. I also removed unused ``import`` statements.

> Are there any other simple but useful tests that we could add? Maybe running experiments via the Python API (= a fork launcher)? I guess we would have to copy the experiments directory to a temporary location for that to work.

I have added some cli tests via ``subprocess.check_call()``. As we only verify the return code with this method we might run into similar problems like the following;
``simex e launch`` will launch the experiments **or** skip experiments if the instance is not available. Both actions will not cause an error and the program ends with return code ``0``.

This means, that we might want to parse the ``print()`` outputs of a subprocess and compare it to the expected output, to avoid different outcomes with the same return code.

avdgrinten (approved):
LGTM.

PR #93, opened 2020-12-08
Support VCS-less builds

### LINKED ISSUE #87: Support builds without VCS
Support builds that do not use git. Note that since revisions need some kind of commit identifier, we can only support git-less builds for dev revisions.

duylethanh:
This PR resolves #87.

The formatting of the changes in ``build.py`` are formatted weirdly. I only encapsulated the current checkout process by ``if build.info.git_repo is not None:`` to check that it is a git-build and ``print("simexpal: Skipping checkout-phase for git-less dev-build {}".format(build.name))`` otherwise. After the checkout process I make sure that the ``/develop/<build_name>@<revision_name>`` directory exists in case the build is git-less.

### DIFF
--- a/docs/builds.rst
+++ b/docs/builds.rst
@@ -14,11 +14,14 @@ only Git is supported) and build them automatically. To achieve this, we need to
 
 On this page we will explain the ``builds`` key and mainly use the
 `C++ example <https://github.com/hu-macsy/simexpal/tree/master/examples/sorting_cpp>`_ from the
-:ref:`QuickStart` guide in order to do so. For example, we will explain how to specify a Git repository, enable
-the automated build support and the directories involved in the build process. Furthermore we will present
-additional options such as pulling Git submodules and setting additional environment variables for the build process.
+:ref:`QuickStart` guide in order to do so. For example, we will explain how to specify a Git repository or
+a local build that does not come from a VCS, enable the automated build support and the directories involved
+in the build process. Furthermore we will present additional options such as pulling Git submodules and setting
+additional environment variables for the build process.
 
 You can find information regarding the ``revisions`` key on the :ref:`Revisions` page.
+
+.. _SpecifyingGitRepository:
 
 Specifying a Git Repository
 ---------------------------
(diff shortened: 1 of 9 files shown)


docs/builds.rst
```diff
@@ -60,6 +63,18 @@ key to ``true`` in order for simexpal to pull the respective files.
 
 If ``recursive-clone`` is not set, it will default to ``false``.
 
+Specifying Local Builds
```
avdgrinten, line 66: Maybe "Local Builds" -> "Builds without Version Control".

docs/builds.rst
```diff
@@ -60,6 +63,18 @@ key to ``true`` in order for simexpal to pull the respective files.
...
+-----------------------
+
+Specifying a local builds work similar to :ref:`SpecifyingGitRepository`. The differences are
+
+- we omit the ``git`` key *and*
+- store the local build files in ``./develop/<build_name>@<revision_name>``.
+
+(As of now local builds are only supported for :ref:`DevRevisions`)
```
avdgrinten, line 74: Drop the parenthesis, this is important. Maybe even make it bold. Explain the rationale.

simexpal/build.py
```diff
@@ -199,81 +199,89 @@ def do_step(step_yml, default_workdir=None):
...
+						'update', '--init', '--recursive'], cwd=checkout_dir)
+
+			did_work = True
+		else:
+			print("simexpal: Skipping checkout-phase for git-less dev-build {}".format(build.name))
+
+	# Make sure that the build source directory exists for git-less dev-builds.
+	if build.info.git_repo is None and build.revision.is_dev_build and not os.path.isdir(build.source_dir):
+		raise RuntimeError(f"The source directory of the build '{build.name}' with revision '{build.revision.name}'"
+							f" does not exist.")
```
avdgrinten, line 284: Why do we check for `build.revision.is_dev_build` here?

simexpal/build.py
```diff
@@ -199,81 +199,89 @@ def do_step(step_yml, default_workdir=None):
...
 
-		did_work = True
+			if build.info.recursive_clone:
+				# Clone submodules recursively
+				subprocess.check_call(['git', 'submodule',
+						'update', '--init', '--recursive'], cwd=checkout_dir)
+
+			did_work = True
+		else:
+			print("simexpal: Skipping checkout-phase for git-less dev-build {}".format(build.name))
```
avdgrinten, line 279: "git-less" -> "VCS-less".

avdgrinten:
This PR should add a test case. (We should aim for test cases of all new feature additions.)

docs/builds.rst
duylethanh: Done. I also changed this formulation at some other parts of the documentation

simexpal/build.py
duylethanh: Done

docs/builds.rst
duylethanh: Done => **As of now local builds are only supported for** Develop Revisions(as link) **as we can not guarantee reproducibility
without some kind of identifier, e.g, a commit hash.**

simexpal/build.py
duylethanh: ``build.source_dir`` is only available to dev-builds so that we have to do this additional check

duylethanh:
PR updated according to change requests. I also added a unit test for VCS-less builds.

avdgrinten (approved):
LGTM.

PR #95, opened 2020-12-16
Download instances from a URL and Git repo, fix broken 'simex i install --overwrite' command

duylethanh:
- Support download of instances from a URL
- Support download of instances from a Git repository 
    - similar to #60 (differences below)
    - does not checkout the repository during the cloning process (``git clone <url> <directory> --no-checkout``)
    - not limited to one item
    - instances will have the same filenames as in the git repository
- Unit tests for both use cases
- Documentation for both features
- Fix broken ``simex i install --overwrite`` command
    - in cases where ``shortname != unique_filename`` the instance was not overwritten


duylethanh:
This PR should also close #60.

avdgrinten (approved):
LGTM.

PR #99, opened 2021-01-23
Improve performance of no ops

### LINKED ISSUE #94: Performance of no-op simex runs
I noticed that just running `simex e` on an empty file takes 300ms on my (decently fast) laptop. This seems to be mainly due to the import of external packages: base.py imports instances.py and that imports `requests`, which takes quite a bit of time. Additionally, importing `jsonschema` ist not for free.

* Only importing instances.py and `requests` when it we need them should be easy.
* For jsonschema, maybe we can cache validation results (however, that would also require us to cache validation error, or, alternatively, disable the caching when there are validation errors).

duylethanh:
This PR fixes #94 and contains the following:

- only import ``instances.py`` and ``requests`` when needed (speedup of ``~70ms`` on my machine, run time from ``~220ms`` to ``~150ms``)
- caching of validation results in ``validation.cache`` to avoid importing``jsonschema``:
    - if the cache entry matches the modification time stamp of the file, the file will not be validated (thus ``jsonschema`` will not be imported, additional speedup of ``~80ms``)
    - otherwise, the file will be validated (up to ``~5ms`` slowdown compared to validation without caching, caused by the additional operations involved in the caching process)
        - if the validation is succesful, the cache entry will be updated
        - otherwise, the cache entry will not be updated (thus, the file will be validated the next time ``simex`` is called)

### DIFF
--- a/simexpal/util.py
+++ b/simexpal/util.py
@@ -7,7 +7,6 @@ import sys
 import yaml
 import json
 from jsonschema import Draft7Validator
-import warnings
 
 def expand_at_params(s, fn, listfn=None):
 	def subfn(m):
@@ -81,7 +80,7 @@ def read_setup_file(setup_file):
 		last_mod = os.fstat(f.fileno()).st_mtime
 	return setup_dict, last_mod
 
-def validate_setup_file(setup_file):
+def validate_setup_file(basedir, setup_file, setup_file_schema):
 	""" Reads, validates and sanitizes the setup file
 	"""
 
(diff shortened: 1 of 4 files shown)


duylethanh:
Sorry, for all the force pushing. This should be the last for now. I forgot to mention that I refactored ``base.config_for_dir()``, which lead to the removal of ``base.read_and_validate_setup()``. As I'd like to acces the ``basedir`` in ``util.validate_setup_file()`` I changed the parameters it receives, explicitly stating ``basedir``. This in turn, made ``base.read_and_validate_setup()`` unnecessary, as it only joined the ``basedir`` with the ``setup_file``.

PR ready to be reviewed.

simexpal/util.py
```diff
@@ -110,33 +113,82 @@ def _validate_dict(dictionary, source, schema_path):
...
-			warnings.warn(msg, DeprecationWarning)
+	writeback_cache = False
+	do_exit = False
+	try:
+		if setup_file_is_valid:
+			writeback_cache = True
+		else:
+			do_exit = True
+	except NameError:
+		pass
```
avdgrinten, line 160: This is not very elegant. This checks whether `setup_file_is_valid` is initialized? Why don't we set the correct default value at the top of the function (or use `None`)?

simexpal/util.py
```diff
@@ -110,33 +113,82 @@ def _validate_dict(dictionary, source, schema_path):
...
+
+	if do_exit:
+		sys.exit(1)
+
+	if setup_file == 'experiments.yml':
+		for exp in setup_file_dict.get('experiments', []):
+			if exp.get('output', None) == 'stdout':
+				msg = "Specifying the stdout path via 'output: stdout' is deprecated and will be removed in future " \
+						"versions. Use 'stdout: out' instead."
+				warnings.warn(msg, DeprecationWarning)
```
avdgrinten, line 184: This `if setup_file ==` is a hack. Move this to a more specific place (e.g., the Config constructor).

avdgrinten (changes requested):

simexpal/util.py
```diff
@@ -78,13 +78,16 @@ def read_setup_file(setup_file):
 
 	with open(setup_file, 'r') as f:
 		setup_dict = yaml.load(f, Loader=YmlLoader)
-	return setup_dict
+		last_mod = os.fstat(f.fileno()).st_mtime
+	return setup_dict, last_mod
 
-def validate_setup_file(setup_file):
+def validate_setup_file(basedir='.', setup_file='experiments.yml', setup_file_schema='experiments.json'):
```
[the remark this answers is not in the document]
duylethanh: Done

simexpal/util.py
duylethanh: I used ``None`` to avoid the ``try/except`` block

simexpal/util.py
duylethanh: Moved this part to ``__init__`` in ``Config``.

duylethanh:
I made the requested changes. I also had to make a few adjustments to the validation unit test:

- remove the assertion for ``instdir`` in the returned ``dict``, as this was moved to the ``Config`` constructor
- delete ``validation.cache`` files before running the validation, as there could already be a cache file from previous (other) tests

PR ready to be reviewed.

avdgrinten:
Can you clarify why we remove the `validation.cache` file in tests? To ensure that the test re-validate even if the file was validated correctly? Or is there any correctness issue without removing the file?

avdgrinten (approved):
LGTM.

duylethanh:
I fixed the merge conflicts. I also added the following commit 4d113c5:

- added ``util.read_json_file()``, which returns a json-dict and the last modification timestamp of the file
- ``_validate_dict()`` in ``util.validate_setup_file()`` now takes a json-dict as parameter (instead of a ``jsonschema`` schema path)
- setup files will also be validated if the corresponding schema was modified

Also:

> 
> 
> Can you clarify why we remove the `validation.cache` file in tests? To ensure that the test re-validate even if the file was validated correctly? Or is there any correctness issue without removing the file?

After thinking about it, the removal of the cache should not be necessary.  If an ``experiments.yml`` was already validated successfully (because a previous test used it) it shouldnt be re-validated again.

If everything is fine with commit 4d113c5, this PR should be ready to be merged.

avdgrinten (approved):

PR #100, opened 2021-02-08
dynamic variations and fix misleading validation errors

### LINKED ISSUE #58: Variation types
AFAICS simexpal variations have a name and several values which need to be written manually as well as `extra_args`. It would be nice to have cases where a variation has a specific type (i.e., int, bool), and a list (or even better a range) of values that the variation can have.
For example, at the moment if I want to execute an experiment with 1, 2, .. x threads I need to create a single variation `thread1`, `thread2`, ... for each number of threads. Wouldn't be better to just have a variation `number of threads` that takes values in [1, 2, ... x] ?
This would simplify the writing of the `experiments.yml` file, and also the creation of an output dataframe (see #57).

duylethanh:

- added properties ``Variant.extra_args`` and ``Variant.environ``
- fix #58
- added @-variable ``@VARIANT_VALUE:<axis_name>@``
- substitute environment variables of variants/experiments, slurm settings (``num_threads``, ...) and extra arguments of instances/variants
- added documentation on ``range`` and ``enum`` variants
- added unit tests for validating ``range`` and ``enum`` variants
- fix bug where validation errors in the ``instances`` key were misleading


duylethanh:
PR updated and ready for review:
- fixed merge conflicts
- fixed bug in ``schemes/experiments.json`` regarding the ``num_nodes``, ``procs_per_node`` and ``num_threads`` key
- fixed bug in ``launch/slurm.py`` regarding the substitution oof the  ``num_nodes``, ``procs_per_node`` and ``num_threads`` key

simexpal/schemes/experiments.json
```diff
@@ -431,16 +437,54 @@
...
+									"type":  "integer",
+									"minimum": 1
+								},
+								"num_threads": {
+									"type": "integer",
+									"minimum": 1
+								},
+								"environ": {"type": "object"},
+								"slurm_args": {"$ref": "#/definitions/str_or_str_list"},
+								"launcher": {"type": "string"}
```
avdgrinten, line 471: It's a bit ugly that we need to repeat these properties for all cases. Can't we handle that more concisely using another case in the `anyOf` or so?

simexpal/launch/slurm.py
```diff
@@ -42,13 +42,16 @@ def _do_submit(self, cfg, experiment, runs):
...
 			specs = {
-				'manifest': common.compile_manifest(locked[0]).yml
+				'manifest': manifest.yml
 			}
 		else:
+			manifests = [manifest.yml]
+			manifests.extend([common.compile_manifest(run).yml for run in locked[1:]])
 			specs = {
-				'manifests': [common.compile_manifest(run).yml for run in locked]
+				'manifests': manifests
```
avdgrinten, line 54: This looks strange to me. If we're submitting multiple runs here, is it guaranteed that they all have the same variants? If not, the code below would break since it always extracts the `VARIANT_VALUE` of the first run.

simexpal/schemes/experiments.json
duylethanh: The best I could figure out, is to move out the definitions of the properties by adding the subschema:

``` JSON
"slurm_settings": {
	"type": "object",
	"properties": {
		"num_nodes": {
			"type": ["integer", "string"],
			"minimum": 1
		},
		"procs_per_node": {
			"type":  ["integer", "string"],
			"minimum": 1
		},
		"num_threads": {
			"type": ["integer", "string"],
			"minimum": 1
		},
               "slurm_args": {"$ref": "#/definitions/str_or_str_list"},
	}
}
```

and adding it to an ``allOf``:

``` JSON
"allOf": [
		{
			"$ref": "#/definitions/slurm_settings"
		},
		{
			"required": ["axis", "range", "steps"],
			"properties": {
				"axis": {"type": "string"},
				"extra_args": {"$ref":  "#/definitions/str_list"},
				"range": {"$ref": "#/definitions/int_list"},
				"steps": {"type": "integer"},
				"num_nodes": {},   // definition is in slurm_settings
				"procs_per_node": {},   // definition is in slurm_settings
				"num_threads": {},   // definition is in slurm_settings
                                "slurm_args": {},    // definition is in slurm_settings
				"environ": {"type": "object"},
				"launcher": {"type": "string"}
			},
			"additionalProperties": false
		}
	]
```

Note that we still have to list the properties in the second ``allOf`` term, as we are using ``"additionalProperties": false``. This would make the schema a bit more concise since with this PR we list those properties 4 times (with the same definition each time) in our schema.

simexpal/launch/slurm.py
duylethanh: Yes, it is guaranteed that the runs have the same variants since we group them by their experiments in ``submit_multiple()`` and experiments are defined by their variations (among their revision and ``ExperimentInfo``-object).

simexpal/schemes/experiments.json
duylethanh: done as described above

duylethanh:
PR ready for review.

simexpal/launch/slurm.py
avdgrinten: Maybe we should still verify this here.

avdgrinten (approved):
LGTM.

duylethanh:
> LGTM, except for the comment below.

PR ready for review.

Added a check for consistent variants.

Example error message:

```
RuntimeError: Can not start experiment 'quick-sort ~ ba-bubble, bs32, num-cores:2 @ main' 
as Slurm job array. The experiment contains inconsistent variants.
```

duylethanh:
Fixed merge conflict in ``docs/at_variables_reference.rst``.

Ready to be reviewed/merged

PR #101, opened 2021-02-22
tighter Slurm integration, documentation, bug fix

### LINKED ISSUE #92: Tighter integration with batch schedulers
Right now, we can use Slurm to launch experiments but we cannot monitor and/or kill experiments through Slurm.

* In `simex e`, query the batch scheduler to determine if job are still alive or not.
* Add a command to kill currently running jobs.

As an implementation strategy, we could store the job IDs of experiments in some file and use that to invoke `squeue` and `scancel`.

duylethanh:
This PR closes #92 and contains the following:

- query Slurm when encountering experiments submitted to or started by Slurm (details below)
- add ``Status.BROKEN`` for experiments that were launched through Slurm and have an unexpected Slurm status
- add ``simex experiments kill`` command (calls ``scancel`` internally)
      - ~``--submitted`` selects all experiments submitted to Slurm~
      - ~``--started`` selects all experiments started by Slurm~
      - uses the same run selection as in purge, launch etc.
      - by default: selects submitted and started experiments
- add documentation on Slurm launcher
- fix bug where the status cache would be written multiple times when using multiple launchers
- fix bug in run ``cli_selects_run()``, where it was not possible to select all runs that failed or were unfinished

Some details regarding the first bullet point:

When experiments are started by the Slurm launcher, the slurm jobid will be stored in the ``.run`` file. When simexpal determines ``submitted`` or ``started`` for a run that was started through Slurm and Slurm has not yet been queried (``Config.slurm_queried == False``), simexpal will query Slurm via ``squeue``. It will then parse the status of all listed jobs to avoid multiple ``squeue``commands and save them to ``Config.slurm_queried_jobs`` (key: Slurm jobid, value: ``simexpal.Status``) and set ``Config.slurm_queried = True``. Slurm statuses will be resolved as follows:
- Slurm status: ``PD`` -> simexpal: ``submitted``
- ``R``, ``CG`` -> ``started``
- any other (unexpected) status -> ``broken``

The parsed output can then be accesses via ~``Config.pop_slurm_queried_jobs(jobid)``~ ``Config.get_slurm_job_status(jobid)``. When a ``jobid`` that was not listed in ``squeue`` is provided, ``failed`` will be returned.

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -701,7 +701,7 @@ def do_experiments_launch(args):
 
 		submit_to_launcher(cfg, launcher_runs)
 
-		cfg.writeback_status_cache()
+	cfg.writeback_status_cache()
 
 experiments_launch_parser = experiments_subcmds.add_parser('launch',
 		parents=[run_selection_parser])
--- a/scripts/simex
+++ b/scripts/simex
@@ -122,7 +122,7 @@ def cli_selects_run(args, run):
 	if args.failed:
 		if not run.get_status().is_negative:
 			return False
-	elif args.unfinished:
+	if args.unfinished:
 		status = run.get_status()
 		if not (status.is_neutral or status == Status.NOT_SUBMITTED):
 			return False
--- a/scripts/simex
+++ b/scripts/simex
@@ -8,6 +8,7 @@ import os
 import argcomplete
 import errno
 import re
+import subprocess
 import sys
 
 import simexpal as extl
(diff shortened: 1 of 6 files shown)


simexpal/base.py
```diff
@@ -419,6 +423,46 @@ def export_successful_experiments(self):
...
+
+	def pop_slurm_queried_jobs(self, jobid):
+		# Returns 'failed' for jobs not listed in self.slurm_queried_jobs.
+		# This only holds for jobs that would receive the status
+		# 'submitted' or 'started' in Run._update_status_cache_dict().
+		status = self.slurm_queried_jobs.pop(jobid, Status.FAILED)
+		if len(self.slurm_queried_jobs) == 0:
+			self.slurm_queried = False
+
+		return status
```
avdgrinten, line 458: This function is a bit strange. Why do we pop the status instead of just returning it?

scripts/simex
```diff
@@ -769,6 +770,51 @@ experiments_print_parser = experiments_subcmds.add_parser('print',
...
+			subprocess.run(scancel_cmd)
+	else:
+		print("No Slurm jobs available")
+
+	cfg.writeback_status_cache()
+
+experiments_kill_parser = experiments_subcmds.add_parser('kill')
+experiments_kill_parser.set_defaults(cmd=do_experiments_kill)
+experiments_kill_parser.add_argument('--submitted', action='store_true', help='Kill pending Slurm jobs')
+experiments_kill_parser.add_argument('--started', action='store_true', help='Kill running Slurm jobs')
```
avdgrinten, line 815: Why do we use a different way to select runs in this function (compared to our usual options as in launch/purge/etc)?

simexpal/base.py
```diff
@@ -1230,6 +1230,15 @@ def internal_name(self):
 			internal_name += '[{}]'.format(self.repetition)
 		return internal_name
 
+	@property
+	def slurm_jobid(self):
+		try:
+			with open(self.aux_file_path('run'), 'r') as f:
+				data = yaml.load(f, Loader=YmlLoader)
+			return data.get('slurm_jobid', None)
```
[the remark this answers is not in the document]
duylethanh: Ah, ``yaml.load()`` will return ``None``. I will catch this case and return ``None`` for the ``slurm_jobid``.

simexpal/base.py
duylethanh: Hm, I think I wanted to track whether the status of a job was already queried or not, but in hindsight it is probably not necessary

scripts/simex
duylethanh: It only makes sense to kill experiments that were ``started`` or ``submitted``. But I can allow the usual options and then filter out every status that is not in ``['started', 'submitted']`` so the run selection interface stays consistent

avdgrinten: Yes, that sounds reasonable.

simexpal/launch/slurm.py
```diff
@@ -103,10 +103,18 @@ def substitute(p):
...
+		out, _ = process.communicate(sbatch_script.encode())  # Assume UTF-8 encoding here.
+		return_msg = out.decode()
+		print(return_msg)
+
 		assert process.returncode == 0
+		assert 'Submitted batch job' in return_msg
 
-		for run in locked:
-			common.create_run_file(run)
+		jobid = return_msg.split()[3]
```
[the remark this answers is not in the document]
duylethanh: done

simexpal/base.py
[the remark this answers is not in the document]

simexpal/base.py

scripts/simex
duylethanh: > It only makes sense to kill experiments that were `started` or `submitted`. But I can allow the usual options and then filter out every status that is not in `['started', 'submitted']` so the run selection interface stays consistent

done

duylethanh:
I updated the PR and the description at the top.
Apart from the review comments, I adapted the documentation for ``simex e kill`` and fixed a bug in ``cli_selects_run()``, where it was not possible to select all runs that failed or were unfinished

Ready for review.

avdgrinten (approved):
LGTM.

PR #103, opened 2021-03-05
handle execve() errors gracefully

### LINKED ISSUE #102: Handle execve() errors gracefully
If simexpal cannot execute an experiment, e.g., because the program is missing, we should still create a `.status` file.

duylethanh:
- fix #102 (the ``.status`` file will be created even if simexpal can not find the executable), this also fixed a bug where experiments were shown as ``started`` when the executable was not found
- added unit test for experiments with missing programs

In the ``launch/common.py`` I mainly wrapped

``` python

child = subprocess.Popen(cmd, cwd=cwd, env=environ,
				stdout=stdout, stderr=stderr)

```

in a ``try/except`` block and put the code below it in the ``else`` case.

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -1251,7 +1251,7 @@ class Run:
 				status = Status.TIMEOUT
 			elif status_dict['signal']:
 				status = Status.KILLED
-			elif status_dict['status'] > 0:
+			elif status_dict['status'] > 0 or status_dict.get('error', None):
 				status = Status.FAILED
 			else:
 				status = Status.FINISHED
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -424,8 +424,26 @@ def invoke_run(manifest):
 	start = time.perf_counter()
 	cwd = (util.expand_at_params(manifest.workdir, substitute)
 			if manifest.workdir is not None else manifest.base_dir)
-	child = subprocess.Popen(cmd, cwd=cwd, env=environ,
-			stdout=stdout, stderr=stderr)
+	try:
+		child = subprocess.Popen(cmd, cwd=cwd, env=environ,
+				stdout=stdout, stderr=stderr)
+	except FileNotFoundError:
+		import traceback
+
+		# Log the error traceback.
+		with open(manifest.aux_file_path('stderr'), 'w') as f:
+			f.write(traceback.format_exc())
+
+		# Create the .status file.
+		status_dict = {'timeout': False, 'walltime': 0,
+				'status': -1, 'signal': None,
+				'error': 'executable_not_found'}
+		with open(manifest.output_file_path('status.tmp'), "w") as f:
+			yaml.dump(status_dict, f)
+		os.rename(manifest.output_file_path('status.tmp'), manifest.output_file_path('status'))
+
+		return
+
 	os.close(stderr)
 	sel = selectors.DefaultSelector()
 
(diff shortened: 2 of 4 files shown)


simexpal/launch/common.py
```diff
@@ -424,65 +424,79 @@ def close(self):
...
+		# Log the error traceback.
+		with open(manifest.aux_file_path('stderr'), 'w') as f:
+			f.write(traceback.format_exc())
+
+		# Create the .status file.
+		status_dict = {'timeout': False, 'walltime': 0, 'status': 1, 'signal': None}
+		with open(manifest.output_file_path('status.tmp'), "w") as f:
+			yaml.dump(status_dict, f)
+		os.rename(manifest.output_file_path('status.tmp'), manifest.output_file_path('status'))
 	else:
```
avdgrinten, line 442: I dislike that the entire code is in the `else` block now. Can't we just return in the error case?

simexpal/launch/common.py
```diff
@@ -424,65 +424,79 @@ def close(self):
...
+				stdout=stdout, stderr=stderr)
+	except FileNotFoundError:
+		import traceback
+
+		# Log the error traceback.
+		with open(manifest.aux_file_path('stderr'), 'w') as f:
+			f.write(traceback.format_exc())
+
+		# Create the .status file.
+		status_dict = {'timeout': False, 'walltime': 0, 'status': 1, 'signal': None}
```
avdgrinten, line 438: Not sure if `status`: 1 is a great way to represent this. Maybe we could set `status` = -1 (which should not be a possible return value for successful executions) and add an attribute such as `error: 'executable_not_found'`?

simexpal/launch/common.py
duylethanh: I agree, thats better.

simexpal/launch/common.py
duylethanh: In ``Run._update_status_cache_dict()`` we check for ``status_dict['status'] > 0`` in order to determine failed runs. Also runs that fail without a timeout or signal have ``status_dict['status']== 1``. To use ``status_dict['status']== -1`` we could use the condition ``status_dict['status']> 0 or status_dict['status']< 0`` or alternatively check for the error entry e.g. ``status_dict['status']> 0 or status_dict['error']``. What do you think?

duylethanh: I set ``status = -1``  and ``error: 'executable_not_found'`` when the executable is missing and chose the latter approach  i.e. the condition ``status_dict['status']> 0 or status_dict['error']`` to check for this case.

duylethanh:
PR updated.

simexpal/base.py
```diff
@@ -1251,7 +1251,7 @@ def _update_status_cache_dict(self):
 				status = Status.TIMEOUT
 			elif status_dict['signal']:
 				status = Status.KILLED
-			elif status_dict['status'] > 0:
+			elif status_dict['status'] > 0 or status_dict['error']:
```
avdgrinten, line 1254: What if `error` does not exist (e.g., because we're running on a .status file that was generated by an older version)?

duylethanh: I changed it to ``status_dict.get('error', None)`` to handle this case.

duylethanh:
PR updated.

avdgrinten (approved):
LGTM now, thank you.

PR #106, opened 2021-03-11
Kill child processes when simexpal is terminated prematurely

### LINKED ISSUE #97: Child processes of simexpal keep running even when simexpal is terminated
While working on issue #81, I noticed that experiments started by simexpal keep running even after simexpal is terminated (prematurely).

We should consider catching those cases to terminate all child processes spawned by simexpal.

duylethanh:

- fix #97

If ``SIGINT`` (CTRL + C) or ``SIGTERM`` is caught, kill running child processes before terminating. Note that we can not handle ``SIGKILL`` this way.

### DIFF
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -3,6 +3,7 @@ import os
 import selectors
 import signal
 import subprocess
+import sys
 import time
 
 import yaml
@@ -462,6 +463,19 @@ def invoke_run(manifest):
 	stderr_writer = LazyWriter(stderr_pipe, manifest.aux_file_path('stderr'))
 	sel.register(stderr_pipe, selectors.EVENT_READ, stderr_writer)
 
+	signal_reader, signal_writer = os.pipe()
+	os.set_blocking(signal_writer, False)
+	signal.set_wakeup_fd(signal_writer)
+
+	# We need to install a signal handler in order for
+	# signal.set_wakeup_fd() to write the signal into fd
+	signal.signal(signal.SIGTERM, lambda *args: None)
+	signal.signal(signal.SIGINT, lambda *args: None)
+
+	sel.register(signal_reader, selectors.EVENT_READ)
+
+	do_exit = False
+
 	# Wait until the run program finishes.
 	while True:
 		if child.poll() is not None:
@@ -479,8 +493,15 @@ def invoke_run(manifest):
 		# Consume any output that might be ready.
 		events = sel.select(timeout=1)
 		for (sk, mask) in events:
-			if not sk.data.progress():
+			if sk.fileobj == signal_reader:
+				sel.unregister(signal_reader)
+				child.kill()
+				do_exit = True  # We also need to terminate the parent process. Otherwise subsequent experiments will be launched.
+			elif not sk.data.progress():
 				sel.unregister(sk.fd)
+
+	os.close(signal_reader)
+	os.close(signal_writer)
 
 	# Consume all remaining output.
 	while True:
@@ -513,3 +534,8 @@ def invoke_run(manifest):
 	with open(manifest.output_file_path('status.tmp'), "w") as f:
 		yaml.dump(status_dict, f)
 	os.rename(manifest.output_file_path('status.tmp'), manifest.output_file_path('status'))
+
+	if do_exit:
+		raise RuntimeError(
+			"simexpal received a termination signal (either SIGINT or SIGTERM) and has terminated "
+			"the current child process")
--- a/simexpal/queuesock.py
+++ b/simexpal/queuesock.py
@@ -32,6 +32,19 @@ class Queue:
 		return "{}/{}[{}]".format(display_name, manifest['instance'], manifest['repetition'])
 
 	def run(self):
+		import signal
+
+		signal_reader, signal_writer = os.pipe()
+		os.set_blocking(signal_writer, False)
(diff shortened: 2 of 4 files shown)


avdgrinten:
I think we should integrate the signal handler into our event loop. This makes sure that all the code runs synchronously, without any surprises (e.g., we do not need to care *when* the signal handler runs).

We can do this by using the Python function `set_wakeup_fd`: create a pipe (`os.pipe()`), set the writer side as the wakeup FD, add the reader side to our `selector`. When the pipe becomes readable, read the signal number and terminate the process.

duylethanh:
PR updated.

duylethanh:
I added missing ``os.close()`` calls on the pipes.

simexpal/launch/common.py
```diff
@@ -453,9 +466,16 @@ def close(self):
 		# Consume any output that might be ready.
 		events = sel.select(timeout=1)
 		for (sk, mask) in events:
+			if sk.fileobj == signal_reader:
+				child.kill()
+				sys.exit(1)
```
avdgrinten, line 471: Instead of calling `exit(1)`, I think we should follow the normal code path to clean up (e.g., to write the status file), and then we should re-raise the signal or raise an exception.

duylethanh: Done

duylethanh:
PR updated

simexpal/launch/common.py
```diff
@@ -453,9 +468,16 @@ def close(self):
 		# Consume any output that might be ready.
 		events = sel.select(timeout=1)
 		for (sk, mask) in events:
-			if not sk.data.progress():
+			if sk.fileobj == signal_reader:
+				sel.unregister(signal_reader)
+				child.kill()
+				do_exit = True
```
avdgrinten, line 474: Why is this needed? After the child is killed, our usual event loop logic should notice that the process terminated and proceed correctly, shouldn't it?

simexpal/launch/common.py
```diff
@@ -453,9 +468,16 @@ def close(self):
 		# Consume any output that might be ready.
 		events = sel.select(timeout=1)
 		for (sk, mask) in events:
-			if not sk.data.progress():
+			if sk.fileobj == signal_reader:
+				sel.unregister(signal_reader)
```
avdgrinten, line 472: Why do we need to unregister `signal_reader` here? It probably makes more sense to just clear the status by reading from the pipe.

simexpal/queuesock.py
```diff
@@ -50,6 +63,13 @@ def run(self):
 					connection = Connection(conn)
 
 					self.selector.register(conn, selectors.EVENT_READ, connection)
+				elif sk.fileobj == signal_reader:
+					if cur_subproc is not None:
+						cur_subproc.terminate()
```
avdgrinten, line 68: Why do we use terminate() and not kill() here (as in launchers/common.py)?

simexpal/queuesock.py
```diff
@@ -50,6 +63,13 @@ def run(self):
...
+					os.close(signal_reader)
+					os.close(signal_writer)
+					self.close()
+					sys.exit(1)
```
avdgrinten, line 72: As in common.py, the usual event loop logic should handle this without any extra code.

simexpal/launch/common.py
duylethanh: We also need to terminate the parent process. Otherwise experiments that where started through the same ``simex e launch`` command will be launched. I added a comment to this line to make it clear

simexpal/queuesock.py
duylethanh: The subprocesses in ``queuesock.py`` are all simexpal processes that terminate accordingly when receiving ``SIGTERM`` or ``SIGINT``.
The subprocesses in ``common.py`` can be arbitrary processes, which could catch ``SIGTERM`` and ``SIGINT`` and keep runing. Thus, we use ``kill()`` in ``common.py``, which sends ``SIGKILL`` to make sure, we terminate the subprocess.

simexpal/queuesock.py
duylethanh: Analogously to ``common.py`` we need to terminate the parent process. I changed up the handling of this case to avoid some duplicate code

simexpal/launch/common.py
duylethanh: We have to unregister the ``signal_reader`` at some point, so I think after receiving the first signal, we no longer need to listen to it

duylethanh:
PR updated and ready for review

avdgrinten:
Ah, now I understand why we exit the program. But I still think the code would be cleaner if we let the event loops finish as usual (and clean up everything properly even in the exit path); afterwards, we can throw an exception to exit the whole program (instead of doing `sys.exit`). That way, higher level code can decide if it wants to exit or continue with the next run or so.

duylethanh:
> Ah, now I understand why we exit the program. But I still think the code would be cleaner if we let the event loops finish as usual (and clean up everything properly even in the exit path); afterwards, we can throw an exception to exit the whole program (instead of doing `sys.exit`). That way, higher level code can decide if it wants to exit or continue with the next run or so.

Instead of ``sys.exit()`` we now raise the following error.

``` python
raise RuntimeError(
    "simexpal received a termination signal (either SIGINT or SIGTERM) and has terminated "
    "the current child process")
```

avdgrinten (approved):
LGTM.

PR #109, opened 2021-04-15
Add at-variables 'BASE_DIR' and 'INSTANCE_DIR'

duylethanh:
- added the @-variables ``BASE_DIR`` and ``INSTANCE_DIR``

 Both at-variables can be used in in
- ``extra_paths`` in the ``builds`` key
- ``args``, ``environ`` and ``workdir`` in the ``regenerate``/``configure``/``compile``/``install`` key in ``builds``
- ``args`` and ``workdir`` in the ``experiments`` key

### DIFF
--- a/docs/at_variables_reference.rst
+++ b/docs/at_variables_reference.rst
@@ -16,9 +16,11 @@ specify ``experiments.py @INSTANCE@`` as experiment arguments.
 
 Below, we list all @-variables and where they can be used.
 
+- ``@BASE_DIR@``: path of the directory containing the ``experiments.yml``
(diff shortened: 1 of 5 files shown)


avdgrinten (approved):
LGTM. What is the exact motivation for this change?

duylethanh:
A user needed to acces those directories in order to run some scripts. Its possible to access those directories by using relative paths e.g. ``@PREFIX_DIR_FOR:<some_build>@/../..`` but thats rather hacky. Adding those @-variables will not do any harm at least

PR #111, opened 2021-05-07
fix misleading status of queue launcher experiments and redirect queue launcher errors

### LINKED ISSUE #108: Misleading status of experiments when queue launcher is closed prematurely
When starting experiments through the queue launcher all statuses are set to ``submitted`` as they are added into the ``requests`` queue. When the queue launcher is closed prematurely (``SIGKILL``, ``CTRL + C``, ...) the experiments remain ``submitted``.

It would be better if the experiments status is ``failed``, similarily to #92 and #101.
As implementation strategy we could use the approach mentioned in #92, i.e. store some kind of job id and query the queue launcher with them to verify their status.

### LINKED ISSUE #110: queue launcher does not create an .err file if an error occurs
As the title says, the queue launcher does not create an ``.err``file, if an error occurs in simexpal.
For example: Using ``@INSTANCE@`` in experiments with fileless instances causes ``RuntimeError: The instance 'foo' is fileless``(assuming the fileless instance is called ``foo``)

We should redirect such errors to an ``.err`` file in the ``aux/_queue/`` directory just as simexpal does it for experiments launched with Slurm.

(Implement this while fixing #108)

duylethanh:
- fixes #108
    - store a ``queue_jobid`` in the ``.run`` file
    - add ``get_job_status_dict`` action to the queue launcher
    - query the queue launcher when determining the status of experiments launched by a queue launcher
- fixes #110 
    - redirect queue launcher errors to an ``/aux/_queue/<queue_jobid>.err`` file

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -1296,6 +1296,16 @@ class Run:
 		except FileNotFoundError:
 			return None
 
+	def get_queue_jobid(self):
+		try:
+			with open(self.aux_file_path('run'), 'r') as f:
+				data = yaml.load(f, Loader=YmlLoader)
+				if data is None:
+					return None
+			return data.get('queue_jobid', None)
+		except FileNotFoundError:
+			return None
+
 	# Contains auxiliary files that SHOULD NOT be necessary to determine the result of the run.
 	def aux_file_path(self, ext):
 		return os.path.join(self.experiment.aux_subdir,
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -11,6 +11,7 @@ import tempfile
 import warnings
 
 from . import util
+from . import queuesock
 
 DEFAULT_DEV_BUILD_NAME = '_dev'
 EXPERIMENTS_LIST_THRESHOLD = 30
@@ -86,6 +87,9 @@ class Config:
 
 		self.slurm_queried = False
 		self.slurm_queried_jobs = {}
+
+		self.queue_queried = False
+		self.queue_queried_jobs = {}
 
 		self._insts = OrderedDict()
 		self._build_infos = OrderedDict()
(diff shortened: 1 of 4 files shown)


duylethanh:
> AFAICT this PR depends on the Slurm PR #101? Or at least, it has some commits from the Slurm PR? I guess it makes sense to wait with a review until #101 is merged.

Fixed merge conflicts after PR #101 has been merged.

PR ready for review.

simexpal/base.py
```diff
@@ -1328,24 +1362,31 @@ def _update_status_cache_dict(self):
...
 				if jobid is not None:
 					if not self._cfg.slurm_queried:
 						self._cfg.query_slurm()
 					status = self._cfg.get_slurm_job_status(jobid)
-
-					return status
+				elif queue_jobid is not None:
+					if not self._cfg.queue_queried:
+						self._cfg.query_queue()
+					status = Status(self._cfg.get_queue_job_status(queue_jobid))
```
fabratu, line 1375: `self._cfg.get_slurm_job_status` is automatically of type Status-class. Is the same possible (without much hassle) for `self._cfg.get_queue_job_status`?

simexpal/base.py
```diff
@@ -1328,24 +1362,31 @@ def _update_status_cache_dict(self):
 			if os.access(self.output_file_path('out'), os.F_OK):
 				status, last_mod = Status.STARTED, os.stat(self.output_file_path('out')).st_mtime
 
+
```
fabratu, line 1365: Nitpicky: line can be removed.

simexpal/base.py
```diff
@@ -1328,24 +1362,31 @@ def _update_status_cache_dict(self):
...
 				if jobid is not None:
 					if not self._cfg.slurm_queried:
 						self._cfg.query_slurm()
 					status = self._cfg.get_slurm_job_status(jobid)
-
-					return status
+				elif queue_jobid is not None:
+					if not self._cfg.queue_queried:
+						self._cfg.query_queue()
+					status = Status(self._cfg.get_queue_job_status(queue_jobid))
```
fabratu, line 1389: Same as above.

simexpal/launch/queue.py
```diff
@@ -27,17 +27,19 @@ def submit(self, cfg, run):
 
 		if not common.lock_run(run):
 			return
-		common.create_run_file(run)
 
 		specfd, specfile = tempfile.mkstemp(prefix='', suffix='-spec.yml',
 											dir=os.path.join(cfg.basedir, 'aux/_queue'))
 
+		queue_jobid = util.extract_file_prefix_from_path(specfile, '-spec')
```
fabratu, line 34: Is `extract_file_prefix_from_path` also returning a valid `queue_jobid`? If not, we should add proper error handling.

simexpal/queuesock.py
```diff
@@ -74,7 +76,15 @@ def run(self):
 
 					if request:
 						if request['action'] == 'launch':
-							requests.append(request)
+							queue_jobid = util.extract_file_prefix_from_path(request['specfile_path'], '-spec')
```
fabratu, line 79: Same here.

fabratu (changes requested):

simexpal/base.py
duylethanh: We could type-cast before returning the value in ``get_queue_job_status()``, but I'm not sure if we can do it without type-casting.

Currently we get the job status via communication from the queue socket. The yaml-dumper used in line ``87`` in ``queuesock.py`` can not handle objects. Thats why I originally type-cast the Status to its ``int`` representation and later cast it back to its ``Status`` representation.

simexpal/base.py
duylethanh: Whoops, overlooked that one.

simexpal/launch/queue.py
duylethanh: Yes, ``extract_file_prefix_from_path()`` reads the queue_jobid from the file created by ``tempfile.mkstemp()`` in the lines before, so it will always be a valid queue_jobid

simexpal/base.py
duylethanh: I type-casted before returning the value in ``get_queue_job_status()`` (and added a comment explaining why it is needed) so the interface is the same for ``get_queue_job_status()`` and ``get_slurm_job_status()``

duylethanh:
PR ready for review

fabratu (approved):
LGTM.

PR #112, opened 2021-05-13
Add progress bar for download of instances

### LINKED ISSUE #105: Add progress bar for downloads
It would be nice to add a progress bar for downloads. This is especially useful when downloading large files.

duylethanh:
- fixes #105 

A progress bar will be displayed when instances are downloaded through the ``simex i install`` command.
It will look like this during the installation

Screenshot 2021-05-13 at 15 22 06

Each ``=`` represents a 2% progress of the download. Currently the unit of the file size is fixed at ``MB``.

### DIFF
--- a/simexpal/instances.py
+++ b/simexpal/instances.py
@@ -1,6 +1,7 @@
 
 import gzip
 import os
+import sys
 import zipfile
 
 from .util import try_mkdir, expand_at_params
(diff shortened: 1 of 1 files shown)


avdgrinten (commented):
Looks good in general.

One thing that could be improved is the appearance when stdout is not a terminal (for example, if you pipe `simex` into `less`: right now we just spam a bunch a progress bars but we could probably remove the `\r` in this case and only draw a new progress bar every 10% or so.

avdgrinten (approved):
LGTM now, thank you.

PR #113, opened 2021-06-15
Allow common 'extra_args' in url instances

duylethanh:
This PR fixes a bug where common ``extra_args`` were not allowed in url instances.


avdgrinten (approved):
LGTM.

PR #114, opened 2021-06-24
Automatic unpacking of URL instances

duylethanh:
- support automatic unpacking of URL isntances via a new ``format`` key
- documentation on automatic unpacking of URL instances
- tests for automatic unpacking of URL instances
- add some missing keys in [experiments.yml reference page](https://simexpal.readthedocs.io/en/latest/experiments_reference.html)

### DIFF
--- a/simexpal/instances.py
+++ b/simexpal/instances.py
@@ -4,7 +4,8 @@ import os
 import sys
 import zipfile
 
-from .util import try_mkdir, expand_at_params
+from .util import try_mkdir, try_rmfile, try_rmtree, expand_at_params
+from .base import ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS
 
 class DownloadException(Exception):
 	pass
@@ -98,8 +99,46 @@ def download_instance(inst_yml, instances_dir, filename, partial_path, ext):
 				raise RuntimeError("Unexpected parameter {}".format(p))
 
 			url = expand_at_params(inst_yml['url'], substitute)
-			download_path = partial_path + ext
+			download_path = os.path.join(instances_dir, filename + '.download')
 			download_as_stream(url, download_path)
+
+			if inst_yml.get('format', None):
+				import shutil
+				import tempfile
+
+				if inst_yml['format'] in ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS:
+					if inst_yml['format'] == 'gz':
+						fd, tmp_file_path = tempfile.mkstemp(dir=instances_dir)
+						with gzip.open(download_path, 'rb') as f_in:
+							with open(tmp_file_path, 'wb') as f_out:
+								shutil.copyfileobj(f_in, f_out)
+
+						os.rename(tmp_file_path, partial_path + ext)
+				else:
+					tmp_dir_path = tempfile.mkdtemp(dir=instances_dir)
+					try:
+						shutil.unpack_archive(download_path, extract_dir=tmp_dir_path, format=inst_yml['format'])
+					except (ValueError, shutil.ReadError):
+						print("Skipping unpacking of instance '{}' with unsupported archive format: .{}".format(
+							shortname, inst_yml['format']))
+						try_rmtree(tmp_dir_path)
+						try_rmfile(download_path)
+						return False
+
+					try:
+						os.rename(os.path.join(tmp_dir_path, filename), partial_path + ext)
+					except FileNotFoundError:
+						print("Skipping unpacking of instance '{}' as it does not exist in the archive '{}'".format(
+							filename, url
+						))
+						try_rmtree(tmp_dir_path)
+						try_rmfile(download_path)
+						return False
+					try_rmtree(tmp_dir_path)
+
+				try_rmfile(download_path)
+			else:
+				os.rename(download_path, partial_path + ext)
 		elif method == 'git':
 			import subprocess
 
--- a/docs/experiments_reference.rst
+++ b/docs/experiments_reference.rst
@@ -41,6 +41,7 @@ used for specifying builds:
 - ``configure``: list of dictionaries containing configuration parameters
 - ``compile``: list of dictionaries containing compilation parameters
 - ``environ``: dictionary of (environment variable, value)-pairs
+- ``extra_paths``: list of extra paths, which simexpal should check when running an experiment that uses this build
 - ``git``: link to the Git repository
 - ``install``: list of dictionaries containing installation parameters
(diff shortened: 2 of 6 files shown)


simexpal/instances.py
```diff
@@ -98,8 +99,31 @@ def substitute(p):
...
+				try:
+					shutil.unpack_archive(download_path, extract_dir=instances_dir, format=inst_yml['format'])
+					os.rename(os.path.join(instances_dir, filename), partial_path + ext)
+				except (ValueError, shutil.ReadError):
+					if inst_yml['format'] in ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS:
+						if inst_yml['format'] == 'gz':
+							with gzip.open(download_path, 'rb') as f_in:
+								with open(tmp_path, 'wb') as f_out:
+									shutil.copyfileobj(f_in, f_out)
+							os.rename(tmp_path, partial_path + ext)
```
avdgrinten, line 118: I don't like the fact that we first try one unpacking function and then fall back to a different one. What's the reason to do that? Why don't we check against `ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS` first and fall back to `shutil.unpack_archive` if we do not find the format in the list?

simexpal/instances.py
```diff
@@ -98,8 +99,31 @@ def substitute(p):
...
-			download_path = partial_path + ext
+			download_path = os.path.join(instances_dir, filename + '.download')
 			download_as_stream(url, download_path)
+
+			if inst_yml.get('format', None):
+				import shutil
+
+				tmp_path = os.path.join(instances_dir, filename + '.tmp')
+				try:
+					shutil.unpack_archive(download_path, extract_dir=instances_dir, format=inst_yml['format'])
```
avdgrinten, line 110: I think we should extract to a temporary subdirectory of instances_dir here. The zip file might contain all kinds of files that clash with other instances.

avdgrinten: We can use `os.rename` afterwards to move the right file to the instances_dir, and remove the temporary directory.

avdgrinten (commented):
Some comment below. A more general question is: how do we control which file we extract from the archive (in case the archive contains multiple files)?

simexpal/instances.py
duylethanh: So we should do something like the following?:
``` python
#key: format; value: function for unpacking 
ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS = {"gz": func_to_extract_gz}  format
```

``` python

if inst_yml['format'] in ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS:
    # use function at ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS[inst_yml['format']] to unpack
else:
    try:
        shutil.unpack_archive(...)
    except appropriate_errors:
         # skip unpacking of this instance
```

simexpal/instances.py
duylethanh: I agree

duylethanh:
> 
> 
> Some comment below. A more general question is: how do we control which file we extract from the archive (in case the archive contains multiple files)?

Maybe we could allow the following syntax motivated by the syntax of [arbitrary input files](https://simexpal.readthedocs.io/en/latest/instances.html#arbitrary-input-files):

``` yaml
instances:
     - method: url
       url: https://some_url/@INSTANCE_FILENAME@.txt.gz
       items:
         - 'bar'
         - name: foo
            files:
              - file1
              - file2
       format: gz       
 ```
 
 In the "normal" case where we just have one file we subtitute ``bar`` into ``@INSTANCE_FILENAME@``, download from that URL and unpack it (according to ``format``) in a temp dir and move the respective file .

In the second case, where we have multiple files we substitute  ``foo`` into ``@INSTANCE_FILENAME@``, download from that URL and unpack it in a temp dir and move the respective files from the ``files`` list.

simexpal/instances.py
avdgrinten: Yes, that sounds better.

duylethanh: I changed it according to above, but did not include the unpacking function into ``ADDITIONAL_SUPPORTED_ARCHIVE_FORMATS``.

simexpal/instances.py
duylethanh: Done.

duylethanh:
> > Some comment below. A more general question is: how do we control which file we extract from the archive (in case the archive contains multiple files)?
> 
> Maybe we could allow the following syntax motivated by the syntax of [arbitrary input files](https://simexpal.readthedocs.io/en/latest/instances.html#arbitrary-input-files):
> 
> ```yaml
> instances:
>      - method: url
>        url: https://some_url/@INSTANCE_FILENAME@.txt.gz
>        items:
>          - 'bar'
>          - name: foo
>             files:
>               - file1
>               - file2
>        format: gz       
> ```
> 
> In the "normal" case where we just have one file we subtitute `bar` into `@INSTANCE_FILENAME@`, download from that URL and unpack it (according to `format`) in a temp dir and move the respective file .
> 
> In the second case, where we have multiple files we substitute `foo` into `@INSTANCE_FILENAME@`, download from that URL and unpack it in a temp dir and move the respective files from the `files` list.

As far as I know, jsonschema can not validate such schemes (where the array has elements of different types at the same time - except if we set the ``items`` key to an arbitrary array without validating its elements, but I'm not sure thats the way to go), so we can dismiss this approach for now.

It does not matter if the archive contains multiple files, if we only want to extract one.
If we want to support the extraction of multiple files, we might be able to support it if we only allow dicts in the ``items`` key when the ``format`` key is present, e.g.:

 ```yaml
 instances:
      - method: url
        url: https://some_url/@INSTANCE_FILENAME@.zip
        items:
          - name: foo
             files:
               - file1
               - file2
           - name: bar
             files:
               - file3
               - file4
        format: zip       
 ```

where ``file1`` and ``file2`` will be extracted from ``foo.zip`` and analogously for ``bar``.

duylethanh:
PR ready to be reviewed

avdgrinten:
To clarify: if the archive is called `Bar.zip` (and stored at a location `https://example.com/Bar.zip`), how would I make simexpal extract file `Foo` out of `Bar.zip`?

duylethanh:
Seems like I did not implement it at that time , but it would probably be like in the comment [above](https://github.com/hu-macsy/simexpal/pull/114#issuecomment-1013743476). I will implement it after I'm done with refining the other open PRs.

bernlu:
Downloading instances is no longer planned for future releases.

PR #116, opened 2021-07-02
Support postprocessing of instances

duylethanh:
- support arbitrary postprocessing of instances by allowing a list of dictionaries containing the ``args``, ``workdir`` and ``environ`` keys in the ``postprocess`` key (like the build arguments in ``compile``, ``configure``, ...)

I will add
- documentation on postprocessing of instances
- tests for postprocessing of instances

before marking the PR as ready


simexpal/base.py
```diff
@@ -693,64 +693,191 @@ def check_available(self):
 	def install(self):
 		from . import instances
 
-		global DID_WARN_KONECT
-
-		if self.check_available() or self.is_fileless:
+		if self.is_fileless:
```
duylethanh, line 696: I moved ``check_available()`` to the later line number 702. In this way we can only postprocess instances, if they are already available (possible reasons: its a local instance or postprocessing was added after downloading the instances)

duylethanh:
PR ready for review

simexpal/base.py
```diff
@@ -693,72 +693,191 @@ def check_available(self):
...
+					partial_path = os.path.join(instance_dir, self.unique_filename)
+					try:
+						print("simexpal: Converting instance '{}' to edge list format".format(self.shortname))
+						instances.convert_to_edgelist(self._inst_yml, partial_path + '.post0', partial_path + '.post1')
+						os.rename(partial_path + '.post0', partial_path + '.original')
+					except RuntimeError as e:
+						print("simexpal: An error occurred while postprocessing the instance '{}': {}".format(
+							self.shortname, e))
+						print("simexpal: Restoring original instance files")
+						os.rename(partial_path + '.post0', partial_path)
```
avdgrinten, line 776: Hmmm, if we restore the instance file to the `post0` file, how do we know if any given instance has already been postprocessed or not?

avdgrinten: The old logic here was: the `post0` files are obtained through download, and the instance files are *always* already postprocessed, right?

simexpal/base.py
```diff
@@ -693,72 +693,191 @@ def check_available(self):
...
+						return
+
+					stage = 1
+					util.touch(os.path.join(instance_dir, self.yml_name) + '.postprocessed')
+					did_work = True
+				else:
+					assert isinstance(self._inst_yml['postprocess'], list)
+
+					import subprocess
+					import shutil
```
avdgrinten, line 787: The indentation level here is quite deep. Can we extract this code into a function?

simexpal/base.py
duylethanh: > Hmmm, if we restore the instance file to the `post0` file, how do we know if any given instance has already been postprocessed or not?

For each instance we create a ``.postprocessed`` file (lines ``781`` and ``868``) when the postprocessing was successful. Also, we print out status messages in the terminal. Although we probably shouldn't assume that it is enough. Maybe I could add an additional check in ``Instance.check_available()`` which looks for the ``.postprocessed`` file for instances that are supposed to be postprocessed. In this way, experiments can't be started on unprocessed instances that were supposed to be postprocessed.

> The old logic here was: the `post0` files are obtained through download, and the instance files are _always_ already postprocessed, right?

Yes. With this PR ``.post0`` files are still obtained through download and instance files are postprocessed depending on the existence of the ``.postprocessed`` file

simexpal/base.py
duylethanh: You mean the code from line ``789`` to ``869``?

duylethanh:
Fixed the merge conflicts for now. When the remarks above are resolved or ok with you I will mark this PR as ready for review again.

simexpal/base.py
duylethanh: I think the line numbers changed due to rebasing. Could you specify which lines we should extract into a function?

simexpal/base.py
fabratu: Concerning `.postprocessed` and `Instance.check_available()`: I think it is enough to inform the user when postprocessing has not been successful. Making successful postprocessing mandatory might lead to unexpected results for users (for example if the experiment does not really rely on it in practice).

scripts/simex
```diff
@@ -302,7 +302,10 @@ def do_instances_install(args):
 
 	for instance in cfg.all_instances():
 		if args.overwrite:
-			util.try_rmfile(os.path.join(cfg.instance_dir(), instance.unique_filename))
+			fullpath = instance.fullpath
+			util.try_rmfile(fullpath)
+			util.try_rmfile(fullpath + '.postprocessed')
+			util.try_rmfile(fullpath + '.original')
```
fabratu, line 308: What is with `.post0` and `.post1`?

fabratu (changes requested):

docs/instances.rst
```diff
@@ -467,6 +468,108 @@ and ``set2``, which contains ``instance2`` and ``instance3``.
...
+
+- ``args``: list of postprocessing arguments
+- ``environ``: dictionary of (environment variable, value)-pairs
+- ``workdir``: path of the working directory
+
+keys.
+
+Assume you want to postprocess the ``facebook_combined`` and ``cit-HepTh`` network from
+`SNAP <https://snap.stanford.edu/data/>`_ using two executables ``postprocess1`` and ``postprocess2``, which
+take the path of the instance as parameter. Also, you have to prepend the path for ``postprocess1`` to the
```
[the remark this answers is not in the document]
duylethanh: This is just to demonstrate the possibility/necessity of adding the appropriate PATH-variable in simexpal in order to find the executable for the postprocessing. There might be cases where the PATH was added beforehand, so that is isn't necessary to do so anymore.

PR #125, opened 2022-08-17
Add wheel-build + upload to PiPy on release tag-push. 

fabratu:
This PR adds:

- Update GHA jobs to match current action- and Python-versions
- New job creates a wheel-build + uploads result to PiPy (only triggered on release-tag push to github). This basically automates the release process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,16 +5,16 @@ on:
       - master
 
 jobs:
-  build:
+  build-and-test:
     strategy:
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,5 +23,49 @@ jobs:
...
+  check-release-tag:
+    name: "Check for new release"
+    runs-on: ubuntu-latest
+    outputs:
+      is-release: ${{ steps.do-check.outputs.is-release }}
+    steps:
+      - name: Check release tag ${{ github.ref }}
+        id: do-check
+        run: |
+          if [[ ${{ github.ref }} =~ ^refs\/tags\/[0-9]+\.[0-9]+(\.[0-9]+)?$ ]]; then
```
avdgrinten, line 40: It's too late to change this for NetworKit, but IMHO adding a `v` in front of the version (for example, v1.0) is a bit nicer.

fabratu: Good point, I thought the same when we introduced the automatic release for NetworKit.

avdgrinten (approved):

PR #127, opened 2022-12-12
Fix fileless instances

Lorax66:
Up to now whenever there was a fileless instance and an experiment with @INSTANCE@ it raised an exception. Now the INSTANCE parameter is replaced with the instance name.

### DIFF
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -347,8 +347,7 @@ def invoke_run(manifest):
 	def substitute(p):
 		if p == 'INSTANCE':
 			if manifest.instance_is_filess:
-				raise RuntimeError(f"The instance '{manifest.instance}' is fileless")
-
+				return manifest.instance_name
 			return manifest.instance_dir + '/' + manifest.instance_name
 		elif p.startswith('INSTANCE:'):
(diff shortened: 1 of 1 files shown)


duylethanh:
This behaviour is intended as a fileless instance is defined by its extra arguments and does not have a file. ``@INSTANCE@`` resolves to the file path of a local/remote instance.

When working with fileless instances we use ``@EXTRA_ARGS@`` in the experiment arguments to parse the extra arguments of the fileless instance.

I see that we currently do not have an extra section for "Experiments with Fileless Instances" stating that one needs to use ``@EXTRA_ARGS@`` in this case (We only somewhat state it in the [Experiments with Local/Remote Instances](https://simexpal.readthedocs.io/en/latest/experiments.html#experiments-with-local-remote-instances)). We could add it below the [Experiments with Arbitrary Input File Instances](https://simexpal.readthedocs.io/en/latest/experiments.html#experiments-with-arbitrary-input-file-instances).

fabratu:
I would second @duylethanh here. We should improve documentation to make clear how to handle fileless instances. In addition the runtime error should also include a small hint in order to cope with the situation.

bernlu:
The behavior is as expected, but better documentation is proposed in #130

PR #131, opened 2023-01-18
Adding requirements.txt for doc-build

fabratu:
`importlib-metadata` just recently updated to version 6.0.0 and `argcomplete` currently pins to a version lower than 6. For RtD-builds, this leads to building errors. 

The additional `requirements.txt` file (pinning `importlib-metadata==5.2.0`) in docs-folder is automatically picked up by the builder and resolves this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ setup(
 	package_data={'simexpal': ['schemes/*.json']},
 	scripts=['scripts/simex'],
 	install_requires=[
+		'importlib-metadata==5.2.0',
 		'argcomplete',
 		'requests',
 		'pyyaml',


docs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+importlib-metadata==5.2.0
```
duylethanh, line 1: Shouldn't we also add this line to the ``install_requires`` list in ``setup.py``?

duylethanh (changes requested):

docs/requirements.txt
fabratu: I did a few test installs - outside of RtD-builds I could not reproduce the issue. However you are probably right.

c-bebop:
Thanks!

duylethanh:
LGTM

duylethanh (approved):

fabratu:
@duylethanh Thx!

PR #132, opened 2023-02-08
Pin importlib-metadata to <5.0 for argcomplete.

fabratu:
Did a bit more testing, this PR should finally fix the unsuccessful Rtd builds. At least it does locally when reproducing the official pipeline from Rtd.


c-bebop:
Thanks!

Issue #141, opened 2023-11-06
Develop Revisions: Update Documenation


c-bebop:
When using the develop revisions option setting `develop: true`, the git commit/branch entries under the key `build_version` for each repository will be ignored. Instead, the latest state of the repository of its default branch will be checked out after cloning.

This behavior is not clearly documented. The following text under https://simexpal.readthedocs.io/en/latest/revisions.html#develop-revisions shall be changed: 

From:
_Values specified in the `build_version` dict will be ignored for develop revisions and SimExPal will clone the latest project files._ 

To:
_Values specified for the build_version key will be ignored for develop revisions. Instead, the latest commit of the default branch will be checked out after cloning the repository._

Further we could add:
_Use git commands within the cloned repositories to switch branches or checkout to another commit._

PR #149, opened 2024-01-05
Clearly document ignored key for build_version

c-bebop:
The build_version key will simply be ignored in develop revisions. This is now documented clearly.

This PR addresses #141

### DIFF
--- a/docs/revisions.rst
+++ b/docs/revisions.rst
@@ -63,7 +63,7 @@ are:
 
 Values specified for the ``build_version`` key will be ignored for develop
 revisions. Instead, the latest commit of the default branch will be checked out
-after cloning the repository.
+after cloning the repository initially.
 
 .. code-block:: YAML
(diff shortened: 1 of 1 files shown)


docs/revisions.rst
```diff
@@ -76,6 +77,8 @@ latest project files.
 
 .. note::
    It is possible to have normal and develop revisions at the same time.
+   Use the regular git commands within the cloned repositories used as 
+   develop revisions to switch branches or checkout to another commit.
```
bernlu, line 81: This seems a little too long and doenst get to the point IMO. Maybe like this instead?
> To switch branches or checkout another commit in develop revisions, first call simexpal to clone the repository. Then use regular git commands within the cloned repositories.

c-bebop: How do you like the updated version?

bernlu (approved):

PR #152, opened 2024-01-09
Fix ReadTheDocs build

fabratu:
Read the docs seems to have added the requirement of a config file. This PR adds this config file and also the requirements for a doc-build (like theme). Successful builds from this branch can be seen here: https://simexpal-doctester.readthedocs.io/en/20240109_fix_rtd/


c-bebop (approved):
Looks good. Thank you!

Issue #126, opened 2022-08-26
.lock files have 000 permissions


avdgrinten:
simexpal currently creates lock files with 000 permissions; this is very annoying when moving directories (among other things) since `mv` complains that it cannot open these files. There is no reason not to create these files either as `r--r--r--` or `rw-rw-r--` instead.

PR #153, opened 2024-01-18
Change lock-file permissions to r--r--r--

nizomovs:
Addresses #126.
The mode `0o444` is in octal representation and sets the file permissions to `r--r--r--`


simexpal/launch/common.py
```diff
@@ -26,7 +26,7 @@ def lock_run(run):
 	# can be considered a bug (or deliberate misuse) and will lead to hard failures.
 	try:
 		lockfd = os.open(run.aux_file_path('lock'),
-				os.O_RDONLY | os.O_CREAT | os.O_EXCL, mode=0)
+				os.O_RDONLY | os.O_CREAT | os.O_EXCL, mode=0o444)
```
bernlu, line 29: The mode should be set to read and write (should be `0o666` instead of `0o444`).

bernlu (changes requested):
It would be nice to have a test case as well:
1. run an experiment
2. delete the aux folder (or just all files produced by simexpal) to confirm this works now.

bernlu:
The test should also confirm that the lock file can be copied.

simexpal/launch/common.py
bernlu: actually 444 is fine for copying as well, we can keep that.

nizomovs:
I've now set the mode to `0o666` which allows the user to read/write. The test tries opening the file with read/write permissions and then deletes (hopefully) all files generated by simexpal

tests/experiments/test_file_permissions.py
```diff
@@ -0,0 +1,65 @@
...
+from simexpal.util import try_rmfile, try_rmtree
+
+file_dir = os.path.abspath(os.path.dirname(__file__))
+
+yml_dirs = ['../../examples/sorting',
+            '../../examples/sorting_fileless',
+           ]
+
+@pytest.mark.parametrize('rel_yml_path', yml_dirs)
+def test_experiments_missing_program(rel_yml_path):
```
c-bebop, line 16: Change test name to what is actually being tested.

tests/experiments/test_file_permissions.py
```diff
@@ -0,0 +1,65 @@
...
+    try_rmtree(os.path.join(cfg.instance_dir()))
+
+    # Delete cache files
+    try_read_write_and_delete(cfg.status_cache_path)
+    try_read_write_and_delete(os.path.join(exp_dir, 'validation.cache'))
+
+def try_read_write_and_delete(path):
+    try:
+        fd = os.open(path, os.O_RDWR)
+    except FileNotFoundError:
```
c-bebop, line 61: The test has to fail in this case and not silently continue.

tests/experiments/test_file_permissions.py
```diff
@@ -0,0 +1,65 @@
...
+        return
+
+    os.close(fd)
+    try_rmfile(path)
```
c-bebop, line 65: Would there be a way to rewrite the `try_rmfile(path)` function to `try_rmfile(fd)` so that the issue of deleting a file even though we do not have the permissions does not occur?

c-bebop (changes requested):

bernlu (approved):

Issue #137, opened 2023-06-14
Undefined variant leads to full combination of instances and other variants


fabratu:

duylethanh:
Could you provide the ``experiments.yml``?

fabratu:
Hm, I should have posted the content right away. I fixed the problem after posting the issue. However I think it was something related to the following entries:

```
matrix:
  include:
  ...
    - experiments: [linkpredict-sig-RW]
      instsets: [citeseer]
      axes: ['embeddings']
      variants: ['b65536-citeseer', 'b131072-citeseer', 'b262144-citeseer']
    - experiments: [linkpredict-sig-RW]
      instsets: [karate]
      axes: ['embeddings']
      variants: ['b65536-karate', 'b131072-karate', 'b262144-karate']
    - experiments: [linkpredict-sig-RW]
      instsets: [cora]
      axes: ['embeddings']
      variants: ['b65536-cora', 'b131072-cora', 'b262144-cora']
    - experiments: [linkpredict-sig-RW]
      instsets: [flickr]
      axes: ['embeddings']
      variants: ['b256-flickr', 'b65536-flickr', 'b131072-flickr', 'b262144-flickr']
    - experiments: [linkpredict-sig-RW]
      instsets: [pubmed]
      axes: ['embeddings']
      variants: ['b65536-pubmed', 'b131072-pubmed', 'b262144-pubmed']
    - experiments: [linkpredict-sig-RW]
      instsets: [com-Youtube]
      axes: ['embeddings']
      variants: ['b65536-com-Youtube', 'b131072-com-Youtube', 'b262144-com-Youtube']
```

One variant had a misspelling, which therefore led to an undefined variant name. As a result, all `instsets` and also all previously defined `variants` were used for this experiment.

meffertj:
I believe this is intended behaviour, take a look at the "run matrix" page: https://simexpal.readthedocs.io/en/latest/run_matrix.html#runmatrix

here it says: 
"Simexpal will then build the cross product of each experiment, instance set, variant and revision for each entry in include. **Omitting a key in include will select all possible elements of the respective key, e.g., omitting revisions is equivalent to specifying the revisions key with all possible revisions.**"

bernlu:
We should check for undefined variant names and print errors instead of silently ignoring them.

Issue #135, opened 2023-04-17
undefined axis in run matrix does not cause an error


bernlu:
Simex does not warn the user about undefined axes in the run matrix. Instead, the runs are just not added to the experiment list at all.

#### correct experiments.yml
```
instances:
  - repo: local
    items:
      - foo-inst

experiments:
  - name: foo
    args: ['@INSTANCE@', '@EXTRA_ARGS@']
    stdout: out

variants:
  - axis: 'ax1'
    items:
      - name: 'v1'
        extra_args: ['-v1']
      - name: 'v2'
        extra_args: ['-v2']

matrix:
  include:
    - experiments: [foo]
      axes: [ax1]
```
```
$ simex e list
Experiment                                    Instance                            Status
----------                                    --------                            ------
foo ~ v1                                      foo-inst                            [0] not submitted
foo ~ v2                                      foo-inst                            [0] not submitted
2 experiments in total
```

#### broken experiments.yml
(added undefined axis `bar` to the run matrix)
```
instances:
  - repo: local
    items:
      - foo-inst

experiments:
  - name: foo
    args: ['@INSTANCE@', '@EXTRA_ARGS@']
    stdout: out

variants:
  - axis: 'ax1'
    items:
      - name: 'v1'
        extra_args: ['-v1']
      - name: 'v2'
        extra_args: ['-v2']

matrix:
  include:
    - experiments: [foo]
      axes: [ax1, bar]
```
```
$ simex e list
Experiment                                    Instance                            Status
----------                                    --------                            ------
0 experiments in total
```

bernlu:
very similar / possible duplicate to #137

PR #154, opened 2024-01-18
Check for undefined variants and axes

nizomovs:
Addresses #135 and #137: when an undefined variant or axes is used in the experiments matrix, an error is thrown
- adds new `Config._axes` field (currently an `OrderedDict` with the value being the key itself)
- adds two simple test experiment setup files with an undefined variant and an undefined axes (identical to `examples/sorting/sorting_cpp` except for minor typos)

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -613,9 +613,10 @@ class Config:
 						return False
 			return True
 
-		for variant in scope.variants:
-			if variant not in self._variants:
-				raise RuntimeError("Variant {} does not exist".format(variant))
+		if scope.variants is not None:
+			for variant in scope.variants:
+				if variant not in self._variants:
+					raise RuntimeError("Variant {} does not exist".format(variant))
 
 		if scope.axes is None:
 			axes = self._axes.keys()
@@ -1495,4 +1496,3 @@ def config_for_dir(basedir=None):
 		basedir = '.'
 	yml = util.validate_setup_file(basedir, 'experiments.yml', 'experiments.json')
 	return Config(os.path.abspath(basedir), yml)
-
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -171,7 +171,7 @@ class Config:
 					raise RuntimeError("The revision name '{}' is ambiguous".format(revision.name))
 				self._revisions[revision.name] = revision
 
-		for axis in sorted(construct_axes(), key=lambda axis: axis):
+		for axis in sorted(construct_axes()):
 			if axis in self._axes:
 				raise RuntimeError("The axis name '{}' is ambiguous".format(axis))
 			self._axes[axis] = axis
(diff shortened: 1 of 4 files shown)


simexpal/base.py
```diff
@@ -165,6 +171,11 @@ def construct_variants():
 					raise RuntimeError("The revision name '{}' is ambiguous".format(revision.name))
 				self._revisions[revision.name] = revision
 
+		for axis in sorted(construct_axes(), key=lambda axis: axis):
```
bernlu, line 174: `key=lambda axis: axis` is the default - why write this explicitly?

tests/experiments/experiments_ymls/undefined_axes/experiments.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: 'bs64'
+        extra_args: ['64']
+
+matrix:
+  include:
+    - experiments: [quick-sort]
+      variants: [ba-insert, bs32]
+      revisions: [main]
+    - experiments: [quick-sort]
+      axes: [block.size]
```
bernlu, line 55: adding a comment here noting that this is the intentional broken line of the `experiments.yml` file might be useful for future reference & debugging (in case something else breaks in this file at some point)

tests/experiments/experiments_ymls/undefined_variant/experiments.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: 'bs64'
+        extra_args: ['64']
+
+matrix:
+  include:
+    - experiments: [quick-sort]
+      variants: [ba-insert, bs32]
+      revisions: [main]
+    - experiments: [quick-sort]
+      variants: [ba-buble]
```
bernlu, line 55: same thought about a comment here

bernlu (changes requested):

simexpal/base.py
nizomovs: I wrote it explicitly to keep that code consistent with the other parts and with the idea of adding an `Axis` class that could f.e. have references to its variants, which I didn't do but then left that code as is. I can remove that redundancy - should I also change `Config._axes` to a list, rather than a dict?

simexpal/base.py
```diff
@@ -600,8 +613,13 @@ def scope_selects_all_variants_of_axis(axis):
 						return False
 			return True
 
+		if scope.variants is not None:
+			for variant in scope.variants:
+				if variant not in self._variants:
+					raise RuntimeError("Variant {} does not exist".format(variant))
```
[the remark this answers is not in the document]
c-bebop: Also don't miss the `.` at the end of the error message.

c-bebop (changes requested):

bernlu (approved):

Issue #143, opened 2023-11-17
Add slurm exclusive flag to experiment


bernlu:
For the slurm scheduler, add support for the `--exclusive` flag. This is a very common use case (experiments with networkit usually want to use all resources a node offers) and I am currently using a workaround by setting `num_threads` to 24 (in our case).
Of course, one could use `slurm_args: ['--exclusive']`, but since this is a common use case for new users, there should be an easy way to set this.

Additionally, we should work on the documentation for the slurm args (https://simexpal.readthedocs.io/en/latest/experiments.html#sbatch-ntasks-per-node-c-n): currently, the explanation is (imo) very unclear and just aimed at expert users. We should point the reader to the exclusive flag and mention that the other configs are for advanced usage or for MPI.

c-bebop:
Add an experiments flag `exclusive: true` using a boolean with default to `false` if not present in the user config.

PR #157, opened 2024-01-29
Add Slurm exclusive flag

nizomovs:
Addresses #143:
- adds new `exclusive` property to the Slurm launch properties
- if `exclusive: true` is set, the `--exclusive` flag is added to the Slurm arguments

### DIFF
--- a/docs/experiments.rst
+++ b/docs/experiments.rst
@@ -329,7 +329,7 @@ the ``experiments.yml``:
 - ``procs_per_node``: number of tasks to invoke on each node (slurm: ``--ntasks-per-node=n``)
 - ``num_threads``:    number of cpus required per task (slurm: ``-c``, ``--cpus-per-task=ncpus``)
 - ``num_nodes``:      number of nodes on which to run (N = min[-max]) (slurm: ``-N``, ``--nodes=N``)
-- ``exclusive``:      boolean to set the ``--exclusive`` flag, so the job allocation reserves all resources of a node
+- ``exclusive``:      boolean flag to run an experiment exclusively on specified computing resources (slurm: ``--exclusive``)
 
 .. code-block:: YAML
    :linenos:
--- a/docs/experiments.rst
+++ b/docs/experiments.rst
@@ -327,8 +327,9 @@ Currently, simexpal supports the following three ``sbatch`` parameters by using 
 the ``experiments.yml``:
 
 - ``procs_per_node``: number of tasks to invoke on each node (slurm: ``--ntasks-per-node=n``)
-- ``num_threads``: number of cpus required per task (slurm: ``-c``, ``--cpus-per-task=ncpus``)
-- ``num_nodes``: number of nodes on which to run (N = min[-max]) (slurm: ``-N``, ``--nodes=N``)
+- ``num_threads``:    number of cpus required per task (slurm: ``-c``, ``--cpus-per-task=ncpus``)
+- ``num_nodes``:      number of nodes on which to run (N = min[-max]) (slurm: ``-N``, ``--nodes=N``)
+- ``exclusive``:      boolean to set the ``--exclusive`` flag, so the job allocation reserves all resources of a node
 
 .. code-block:: YAML
    :linenos:
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -1213,20 +1213,7 @@ class Experiment:
 		return self.info.thread_settings
 
 	@property
-	def is_effective_exclusive(self):
-		s = None
-		for variant in self.variation:
-			vs = variant.is_exclusive
-			if vs is None:
-				continue
-			if s is not None and s.is_exclusive != vs:
-				raise RuntimeError("Exclusive flag for experiments '{}' overridden by multiple variants: {}".format(
-					self.name, [s.name, variant.name]
-				))
-			s = variant
-
-		if s is not None:
-			return s.is_exclusive
+	def is_exclusive(self):
 		return self.info.is_exclusive
 
 	@property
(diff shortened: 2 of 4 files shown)


simexpal/launch/slurm.py
```diff
@@ -125,12 +125,15 @@ def _represents_int(x):
 		sbatch_args = ['sbatch', '-J', experiment.display_name]
 		if self.queue:
 			sbatch_args += ['-p', self.queue]
-		if ps and ps['num_nodes']:
-			sbatch_args += ['-N', str(ps['num_nodes'])]
+		if not experiment.is_effective_exclusive:
```
c-bebop, line 128: I think it's easier to read if you revert the logic here as in:
```python
if experiment.is_effective_exclusive:
    sbatch_args += ['--exclusive']
else:
    if ps and ps['num_nodes']:
        sbatch_args += ['-N', str(ps['num_nodes'])]
    if ts and ts['num_threads']:
        sbatch_args += ['-c', str(ts['num_threads'])]
```

simexpal/base.py
```diff
@@ -1204,6 +1212,23 @@ def effective_thread_settings(self):
 			return s.thread_settings
 		return self.info.thread_settings
 
+	@property
+	def is_effective_exclusive(self):
```
c-bebop, line 1216: I don't quite understand the logic of this function. As I understand, this function shall return `True` if the experiment is setup to use computing resources exclusively. 

Now you iterate over all variants (why variants?) and then you check if a variant is exclusive: if it is, you continue, if it is not exclusive but no variant has been stored into `s` it'll also continue with the next element, and if the next element is also not exclusive you raise a `RuntimeError` with both the last and the current variant name (`s.name, variant.name`)?

I would be happy for some clarity on what you are doing here to understand how we can simplify this logic.

c-bebop (changes requested):

simexpal/base.py
nizomovs: The code itself is similar to the `effective_process_settings` and `effective_thread_settings` properties defined before it.

To my understanding, the logic is as follows:
- iterate over all variants, check their "settings" (be it process, threads or exclusiveness)
- if two variants have settings set, throw an Error
- if only one variant has settings set, return that
- if no variant has settings set, return the settings from the experiment

However, the code in `is_effective_exclusive` did indeed contain a bug and wasn't working as intended. I've fixed that and further adapted it: the logic is still similar to the above but the error is only thrown if the exclusiveness of two variants varies.

Furthermore, I think the code for `effective_process_settings` and `effective_thread_settings` doesn't account for the fact that the settings of two variants could be the same, in which case there would be no need to throw an error (like I do now)

nizomovs: And the schema has been changed to allow `num_nodes` and `exclusive: true` to be set.

simexpal/base.py
```diff
@@ -1204,6 +1212,23 @@ def effective_thread_settings(self):
...
+			vs = variant.is_exclusive
+			if vs is None:
+				continue
+			if s is not None and s.is_exclusive != vs:
+				raise RuntimeError("Exclusive flag for experiments '{}' overridden by multiple variants: {}".format(
+					self.name, [s.name, variant.name]
+				))
+			s = variant
+
+		if s is not None:
```
c-bebop, line 1228: The final check can be put into a single line and return statement:
```python
return s.is_exclusive if s is not None else self.info.is_exclusive
```

c-bebop (changes requested):

simexpal/base.py
nizomovs: Simplified it for the `effective_process_settings` and `effective_thread_settings` functions. (Since it is no longer needed for `is_exclusive`)

simexpal/schemes/experiments.json
```diff
@@ -40,21 +40,40 @@
...
-				},
-				"num_threads": {
-					"type": ["integer", "string"],
-					"minimum": 1
+			"allOf": [
+				{
+					"not": {
+						"anyOf": [
+							{
+								"required": ["exclusive", "num_threads"],
```
c-bebop, line 48: We do not want to limit the user to either use `exclusive` or `num_threads`. Both can be set.

simexpal/launch/slurm.py
```diff
@@ -125,12 +125,15 @@ def _represents_int(x):
 		sbatch_args = ['sbatch', '-J', experiment.display_name]
 		if self.queue:
 			sbatch_args += ['-p', self.queue]
+		if experiment.is_exclusive:
+			sbatch_args += ['--exclusive']
+		else:
```
c-bebop, line 130: This therefore (see comment on `exclusive` and `num_threads`) must not be in an else branch. We want to set `num_threads` in any way.

docs/experiments.rst
```diff
@@ -327,8 +327,9 @@ Currently, simexpal supports the following three ``sbatch`` parameters by using
 the ``experiments.yml``:
 
 - ``procs_per_node``: number of tasks to invoke on each node (slurm: ``--ntasks-per-node=n``)
-- ``num_threads``: number of cpus required per task (slurm: ``-c``, ``--cpus-per-task=ncpus``)
-- ``num_nodes``: number of nodes on which to run (N = min[-max]) (slurm: ``-N``, ``--nodes=N``)
+- ``num_threads``:    number of cpus required per task (slurm: ``-c``, ``--cpus-per-task=ncpus``)
+- ``num_nodes``:      number of nodes on which to run (N = min[-max]) (slurm: ``-N``, ``--nodes=N``)
+- ``exclusive``:      boolean to set the ``--exclusive`` flag, so the job allocation reserves all resources of a node
```
c-bebop, line 332: If would recommend a phrasing such as: Boolean flag to run an experiment exclusively on specified computing resources. For Slurm, this means we set the ``--exclusive`` flag.

c-bebop (changes requested):

bernlu (approved):

PR #158, opened 2024-01-29
Repo subdir

nizomovs:
Addresses #124:
- adds `subdir` property for `repo` instances

### DIFF
--- a/docs/instances.rst
+++ b/docs/instances.rst
@@ -38,7 +38,7 @@ Local Instances
 ---------------
 
 To add local instances to the ``instances`` key, we add a list of dictionaries with two keys
-to its value:
+and an optional third key to its value:
 
 - ``repo``:   source of the instances
 - ``subdir``: subdirectory of the instances in the ``instdir`` directory
--- a/docs/instances.rst
+++ b/docs/instances.rst
@@ -40,16 +40,31 @@ Local Instances
 To add local instances to the ``instances`` key, we add a list of dictionaries with two keys
 to its value:
 
-- ``repo``: source of the instances
-- ``items``: a list of instances.
+- ``repo``:   source of the instances
+- ``subdir``: subdirectory of the instances in the ``instdir`` directory
+- ``items``:  a list of instances.
 
 An example of how to list a local set of instances is:
 
 .. literalinclude:: ./experiments.yml.example
    :linenos:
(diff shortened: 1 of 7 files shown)


simexpal/schemes/experiments.json
```diff
@@ -171,6 +171,7 @@
 								"extensions": {"$ref": "#/definitions/str_list"},
 								"extra_args": {"$ref": "#/definitions/str_list"},
 								"set": {"$ref": "#/definitions/str_or_str_list"},
+								"subdir": {"type": "string"},
```
c-bebop, line 174: Please document that this `subdir` is now available for in our [instances readthedocs documentation](https://simexpal.readthedocs.io/en/latest/instances.html).

simexpal/base.py
```diff
@@ -750,13 +750,17 @@ def repo_name(self):
...
+	@property
+	def instance_dir(self):
+		return os.path.join(self._cfg.instance_dir(), self._inst_yml.get('subdir', ''))
+
 	@property
 	def git_subdir(self):
 		return self._inst_yml.get('git_subdir', None)
 
 	def check_available(self):
 		for file in self.filenames:
```
c-bebop, line 762: Here you should be able to do something like:
```python
files_available = [os.path.isfile(os.path.join(self.instance_dir, file)) for file in self.filenames]
return all(files_available)
```

c-bebop (changes requested):

c-bebop (dismissed):

docs/instances.rst
```diff
@@ -40,17 +40,32 @@ Local Instances
 To add local instances to the ``instances`` key, we add a list of dictionaries with two keys
```
bernlu, line 40: two keys and an optional third one

bernlu (changes requested):
do we have tests that check the changes in base.py? I am not an expert on that part of the project and cannot review those changes for correctness.

docs/instances.rst
nizomovs: I've fixed the docs entry.

There's also a new test to check the functionality of the `instdir` key and the instance `subdir` keys. The instance files are just empty files and the test checks if all instances are available.

bernlu (approved):

c-bebop (approved):

Issue #142, opened 2023-11-06
Develop Revisions: Change Behavior & CLI


c-bebop:
# Old 

The current implementation of the develop revisions is however not very intuitive and lacks functionality to actually be able to conveniently develop within multiple repositories, and then test or benchmark the current state within a SimExPal environment.

Therefore, I hereby propose the following functionality change:

SimExPal will first clone and then check out the latest commit present in the given branch provided as the key under `build_version` for each repository.

The CLI command `simex develop --checkout` checks the current status of each repository and does the following:
1. If the repository has no local changes: checkout to the given branch and the latest commit (`HEAD`).
3. If the repository has local changes: disregard local changes and checkout to the given branch and the latest commit (`HEAD`). The question here is, if it is enough to simply overwrite these changes and not warn the user that data can be lost before executing this command. Therefore, I recommend doing that only if the `-f` force option is active.

Currently the option `simex develop --recheckout` offers to simply re-clone the repository including a complete re-build and re-configure step. We could also think of changing this command to `simex develop --reclone` to have clearly distinguished commands to checkout or actually re-clone repositories. 

# Update

Our first approach did not succeed and we needed to revert the merge of the feature: https://github.com/hu-macsy/simexpal/pull/165

The issue is that simexpal writes cache files to the project folders. This however leads to an unclean repository state and a pull or checkout can not simply proceed. We must therefore first revisit the issue of cache files and how to store them on root level in one file, instead of storing them in each build folder.

PR #161, opened 2024-01-30
Develop revisions

nizomovs:
Addresses #142. The develop revisions checkout logic is now as follows:

1. if the repository has not been cloned yet:
    1. run `git clone <repo_url>`
    2. run `git checkout <branch_name>`, where `branch_name` is the `build_version` specified in the `experiments.yml`
2. if the repository has already been cloned:
    1. if there are local changes using (checked using `git status --porcelain`):
        1. if the force/confirmation flag (`-f`) has been set: checkout given branch by running `git checkout <branch_name>`
        2. if the force/confirmation flag has not been set: warn the user of local changes and refer to the force/confirmation flag
    2. if there are no local changes: checkout given branch by running `git checkout <branch_name>`

### DIFF
--- a/simexpal/build.py
+++ b/simexpal/build.py
@@ -180,6 +180,25 @@ def make_build_in_order(cfg, build, wanted_builds, wanted_phases, reclone):
 	def log_phase(step):
 		print("simexpal: Running {}-phase for build {}".format(step, build.name))
 
+	def do_develop_checkout(build, reclone):
+		branch = build.revision.version_for_build(build.name)
+		if os.path.isdir(build.source_dir) and not reclone:
+			# Repository already cloned, check for local changes
+			status = subprocess.check_output(['git', 'status', '--porcelain'], cwd=build.source_dir)
+			if len(status) != 0:
+				print("There are local changes that have not yet been commited and pushed. Please clean them up before proceeding.")
+				return
+
+			subprocess.check_call(['git', 'checkout', branch], cwd=build.source_dir)
+			subprocess.check_call(['git', 'pull'], cwd=build.source_dir)
+		else:
+			# Clone the branch of the repository into the build.source_dir
+			util.try_rmtree(build.source_dir)
+			util.try_mkdir(build.source_dir)
+
+			subprocess.check_call(['git', 'clone', build.info.git_repo, build.source_dir])
+			subprocess.check_call(['git', 'checkout', branch], cwd=build.source_dir)
+
 	did_work = False
 
 	def do_step(step_yml, default_workdir=None):
@@ -264,29 +283,7 @@ def make_build_in_order(cfg, build, wanted_builds, wanted_phases, reclone):
 						build.clone_dir,
 						generic_tag])
 			else:
-				branch = build.revision.version_for_build(build.name)
-				if os.path.isdir(build.source_dir) and not reclone:
-					# Repository already cloned, check for local changes
-					status = subprocess.check_output(['git', 'status', '--porcelain'], cwd=build.source_dir)
-					if len(status) != 0:
-						print("There are local changes that have not yet been commited and pushed. Please clean them up before proceeding.")
-						return
-
-					# or non-pushed commits
-					unpushed_commits = subprocess.check_output(['git', 'log', '--branches', '--not', '--remotes'], cwd=build.source_dir)
-					if len(unpushed_commits) != 0:
-						print("There are local commits that have not yet been pushed. Please clean them up before proceeding.")
-						return
-
-					subprocess.check_call(['git', 'checkout', branch], cwd=build.source_dir)
-					subprocess.check_call(['git', 'pull'], cwd=build.source_dir)
-				else:
-					# Clone the branch of the repository into the build.source_dir
-					util.try_rmtree(build.source_dir)
-					util.try_mkdir(build.source_dir)
-
-					subprocess.check_call(['git', 'clone', build.info.git_repo, build.source_dir])
-					subprocess.check_call(['git', 'checkout', branch], cwd=build.source_dir)
+				do_develop_checkout(build, reclone)
 
 			util.touch(os.path.join(checkout_dir, 'checkedout.simexpal'))
 
--- a/simexpal/build.py
+++ b/simexpal/build.py
@@ -5,13 +5,13 @@ import subprocess
 
 from . import util
 
-def make_builds(args, cfg, revision, infos, wanted_builds, wanted_phases):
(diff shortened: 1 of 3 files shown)


simexpal/build.py
```diff
@@ -264,12 +264,27 @@ def git_ref_changed():
 						build.clone_dir,
 						generic_tag])
 			else:
-				# Recreate the source directory
-				util.try_rmtree(build.source_dir)
-				util.try_mkdir(build.source_dir)
+				branch = build.revision.version_for_build(build.name)
+				if os.path.isdir(build.source_dir):
+					# Repository already cloned, check for local changes
+					status = subprocess.check_output(['git', 'status', '--porcelain'], cwd=build.source_dir)
```
c-bebop, line 270: What you need to check here is: am I still on the branch given by the config? If not, and there are local changes: print error message, and do not proceed. If there are no local changes, checkout to configured branch. Now is the question: are there any local commits not pushed? If so: print error message, do not proceed. If everything is fine: pull in the latest changes from origin.

simexpal/build.py
```diff
@@ -65,7 +65,7 @@ class Phase(IntEnum):
 	COMPILE = 4
 	INSTALL = 5
 
-def make_build_in_order(cfg, build, wanted_builds, wanted_phases):
+def make_build_in_order(args, cfg, build, wanted_builds, wanted_phases):
```
c-bebop, line 68: Make a flag out of `args.f` so you don't pass in this external dependency.

c-bebop (changes requested):

simexpal/build.py
nizomovs: The logic is now as follows:

1. if the repository has not been cloned yet:
    1. run `git clone <repo_url>`
    2. run `git checkout <branch_name>`, where `branch_name` is the `build_version` specified in the `experiments.yml`
2. if the repository has already been cloned:
    1. if there are local changes using (checked using `git status --porcelain`) or unpushed commits (checked using `git log --branches --not --remotes`): warn the user of local changes and do not proceed
    2. if there are no local changes and no unpushed commits: checkout given branch by running `git checkout <branch_name>` and then `git pull`

If the `--reclone` flag was set, the installed repository is deleted and recloned, meaning that all the following build phases are executed, as well.

simexpal/build.py
nizomovs: Done. Instead of `args.f`, only the `--reclone` flag (`args.reclone` is passed into the functions, with a default value of `False`)

simexpal/build.py
```diff
@@ -264,12 +264,29 @@ def git_ref_changed():
...
-				# Recreate the source directory
-				util.try_rmtree(build.source_dir)
-				util.try_mkdir(build.source_dir)
-
-				# Clone the git repository into the build.source_dir
-				subprocess.check_call(['git', 'clone', build.info.git_repo, build.source_dir])
+				branch = build.revision.version_for_build(build.name)
+				if os.path.isdir(build.source_dir) and not reclone:
+					# Repository already cloned, check for local changes
+					status = subprocess.check_output(['git', 'status', '--porcelain'], cwd=build.source_dir)
```
c-bebop, line 270: But now you're trying to check out to the branch given as the value for the branch, never mind if you're already on the branch? If so, this should be changed. If you're already on the branch then all these checks don't have to be proceeded with.

simexpal/build.py
```diff
@@ -264,12 +264,29 @@ def git_ref_changed():
...
+					if len(status) != 0:
+						print("There are local changes that have not yet been commited and pushed. Please clean them up before proceeding.")
+						return
+
+					# or non-pushed commits
+					unpushed_commits = subprocess.check_output(['git', 'log', '--branches', '--not', '--remotes'], cwd=build.source_dir)
```
c-bebop, line 276: After reviewing this, I don't think we should force that all changes are pushed. Please leave out checking if all changes have been pushed. Only local changes that are not committed should lead to an error.

simexpal/build.py
```diff
@@ -264,12 +264,29 @@ def git_ref_changed():
 						build.clone_dir,
 						generic_tag])
 			else:
-				# Recreate the source directory
-				util.try_rmtree(build.source_dir)
-				util.try_mkdir(build.source_dir)
-
-				# Clone the git repository into the build.source_dir
-				subprocess.check_call(['git', 'clone', build.info.git_repo, build.source_dir])
+				branch = build.revision.version_for_build(build.name)
```
c-bebop, line 267: This is already a huge function. Please use a dedicated function to implement this logic.

c-bebop (changes requested):

simexpal/build.py
```diff
@@ -264,12 +264,29 @@ def git_ref_changed():
...
+				if os.path.isdir(build.source_dir) and not reclone:
+					# Repository already cloned, check for local changes
+					status = subprocess.check_output(['git', 'status', '--porcelain'], cwd=build.source_dir)
+					if len(status) != 0:
+						print("There are local changes that have not yet been commited and pushed. Please clean them up before proceeding.")
+						return
+
+					# or non-pushed commits
+					unpushed_commits = subprocess.check_output(['git', 'log', '--branches', '--not', '--remotes'], cwd=build.source_dir)
+					if len(unpushed_commits) != 0:
```
c-bebop, line 277: I would say it is the best to not care about the unpushed commits. So this `if` clause can be removed.

simexpal/build.py
```diff
@@ -264,12 +264,29 @@ def git_ref_changed():
...
+						print("There are local commits that have not yet been pushed. Please clean them up before proceeding.")
+						return
+
+					subprocess.check_call(['git', 'checkout', branch], cwd=build.source_dir)
+					subprocess.check_call(['git', 'pull'], cwd=build.source_dir)
```
c-bebop, line 282: Can you check if this pull request was successful? And if not, simply print that? I am a bit hesitant of the first design with pulling from `origin`. What if we're not connected to the required network to resolve the URL? If there is a solution to pull and warn the user if pull was not successful I think that would be the preferred version.

c-bebop (changes requested):

simexpal/build.py
nizomovs: I've removed the check for unpushed commits.

simexpal/build.py
nizomovs: The logic has been moved to a new `do_develop_checkout` function

c-bebop (approved):
Looks good!

PR #172, opened 2024-03-20
Add INSTANCE_FILENAME @-Variable

nizomovs:
The `INSTANCE_FILENAME` @-Variable was only usable for instance generators. However, it is useful in the `extra_args` member of experiments to i.e. allow using custom output directories in the output folder per instance.

### DIFF
--- a/simexpal/launch/common.py
+++ b/simexpal/launch/common.py
@@ -353,9 +353,7 @@ def invoke_run(manifest):
 		elif p.startswith('INSTANCE:'):
 			return get_qualified_filename(p.split(':')[1])
 		elif p == 'INSTANCE_FILENAME':
-			if manifest.instance_is_filess:
-				raise RuntimeError(f"The instance '{manifest.instance}' is fileless. Did you forget to remove the @INSTANCE_FILENAME@ variable in the argument list of the experiment?")
-
+			# @INSTANCE_FILENAME@ is only the instance name and hence can still be used for fileless instances
 			return manifest.instance_name
 		elif p == 'REPETITION':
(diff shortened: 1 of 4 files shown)


simexpal/launch/common.py
```diff
@@ -352,6 +352,11 @@ def substitute(p):
 			return manifest.instance_dir + '/' + manifest.instance_name
 		elif p.startswith('INSTANCE:'):
 			return get_qualified_filename(p.split(':')[1])
+		elif p == 'INSTANCE_FILENAME':
+			if manifest.instance_is_filess:
+				raise RuntimeError(f"The instance '{manifest.instance}' is fileless. Did you forget to remove the @INSTANCE_FILENAME@ variable in the argument list of the experiment?")
```
c-bebop, line 357: I like this error message. Can you make a test for this functionality?

c-bebop (changes requested):

bernlu (dismissed):
Another test that actually uses the new variable would be nice as well.

simexpal/launch/common.py
```diff
@@ -352,6 +352,9 @@ def substitute(p):
 			return manifest.instance_dir + '/' + manifest.instance_name
 		elif p.startswith('INSTANCE:'):
 			return get_qualified_filename(p.split(':')[1])
+		elif p == 'INSTANCE_FILENAME':
```
c-bebop, line 355: Still lacking a test that `@INSTANCE_FILENAME@` can actually be used.

c-bebop (changes requested):

c-bebop (approved):

Issue #169, opened 2024-02-07
Cache Files


c-bebop:
simexpal uses cache files such as `checkedout.simexpal` and `regenerated.simexpal` at the level of each build folder (e.g. `develop/benchmark` for a build `benchmark`).

Having these cache files changes the resource (the build). The goal of this issue is to collectively store all cache information at root level of the simexpal project.

First, we should clarify why exactly cache files are written to the build director? Second, change that behaviour and store all caching information on root level. This could be done using a `.yml` or `.json` file (see `status.cache` and `validation.cache` at root level).

PR #177, opened 2024-07-03
refactor build cache

PetholzA:
This PR refactors the build cache (which previously consisted of `status.cache`, `validation.cache` and various `.simexpal` files) into a unified caching structure referring to #169.

### DIFF
--- a/scripts/simex
+++ b/scripts/simex
@@ -23,6 +23,7 @@ from simexpal.base import Status
 from simexpal.base import YmlLoader
 from itertools import zip_longest
 import yaml
+import json
 import warnings
 
 warnings.simplefilter(action='default', category=DeprecationWarning)  # do not ignore DeprecationWarnings
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -82,7 +82,7 @@ class Config:
 		self.basedir = basedir
 		self.yml = yml
 
-		self.status_cache_path = basedir + '/status.cache'
+		self.status_cache_path = basedir + util.SIMEX_CACHE
 		self.status_cache_dict = {}
 
 		self.slurm_queried = False
@@ -195,8 +195,8 @@ class Config:
 				self._exp_infos[exp_yml['name']] = ExperimentInfo(self, exp_yml)
 
 		try:
-			with open(self.status_cache_path, 'r') as f:
-				self.status_cache_dict = json.load(f)
+			with open(self.status_cache_path, 'r') as cachefile:
+				self.status_cache_dict = json.load(cachefile)["status"]
 		except FileNotFoundError:
 			pass
 
@@ -503,10 +503,11 @@ class Config:
 		return Status(self.queue_queried_jobs.get(jobid, Status.FAILED))
 
 	def writeback_status_cache(self):
-		fd, path = tempfile.mkstemp(dir=self.basedir)
-		with os.fdopen(fd, 'w') as tmp:
-			json.dump(self.status_cache_dict, tmp)
-		os.rename(path, self.status_cache_path)
+		with open(os.path.join(self.basedir, util.SIMEX_CACHE), 'r') as cachefile:
+			cache = json.load(cachefile)
+		cache["status"] = self.status_cache_dict
+		with open(os.path.join(self.basedir, util.SIMEX_CACHE), 'w') as cachefile:
+			json.dump(cache, cachefile)
 
 	# -----------------------------------------------------------------------------------
 	# Matrix expansion.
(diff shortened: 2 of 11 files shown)


scripts/simex
```diff
@@ -394,6 +395,13 @@ builds_remake_parser.set_defaults(cmd=do_builds_remake)
 # ---------------------------------------------------------------------------------------
 
 def do_develop(args):
+	# Initialize the cache.
+	if not os.path.isfile("./simex.cache"):
+		with open('./simex.cache', 'a+') as myfile:
```
c-bebop, line 400: I would think it would be beneficial to have the `simex.cache` file hidden, thus naming it `.simex.cache`. I also think that the name of the cache file should be parameterized as a constant. This way, the naming of important files is not _hidden_ in code branches.

simexpal/base.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, basedir, yml):
 		self.basedir = basedir
 		self.yml = yml
 
-		self.status_cache_path = basedir + '/status.cache'
+		self.status_cache_path = basedir + '/simex.cache'
```
c-bebop, line 85: See comment on file name constants as a variable/parameter.

scripts/simex
```diff
@@ -394,6 +395,13 @@ builds_remake_parser.set_defaults(cmd=do_builds_remake)
 # ---------------------------------------------------------------------------------------
 
 def do_develop(args):
+	# Initialize the cache.
```
c-bebop, line 398: Looking at the size of this function I recommend moving this to a dedicated function instead of adding a comment:
```python
if not os.path.isfile("./simex.cache"):
    init_cache()
```

simexpal/base.py
```diff
@@ -503,10 +503,15 @@ def get_queue_job_status(self, jobid):
...
-		fd, path = tempfile.mkstemp(dir=self.basedir)
-		with os.fdopen(fd, 'w') as tmp:
-			json.dump(self.status_cache_dict, tmp)
-		os.rename(path, self.status_cache_path)
+		#fd, path = tempfile.mkstemp(dir=self.basedir)
+		#with os.fdopen(fd, 'w') as tmp:
+		#	json.dump(self.status_cache_dict, tmp)
+		#os.rename(path, self.status_cache_path)
+		with open(self.basedir + '/simex.cache', 'r') as myfile:
+			cache = json.load(myfile)
```
c-bebop, line 511: Rename `myfile` to `cache_file`?

simexpal/base.py
```diff
@@ -503,10 +503,15 @@ def get_queue_job_status(self, jobid):
...
+			cache["status"] = self.status_cache_dict
+		with open(self.basedir + '/simex.cache', 'w') as myfile:
```
c-bebop, line 513: Why do we need to open the file once with read and then with write access? Why not open it with read write access, load the cache from file, write the cache from memory to the read object `cache` object from file and then write it back to file? Or am I not following what happens here?

simexpal/base.py
```diff
@@ -1013,22 +1018,32 @@ def source_dir(self):
...
-			return os.access(os.path.join(self.source_dir, 'regenerated.simexpal'), os.F_OK)
+			with open(self._cfg.basedir + '/simex.cache', 'r') as myfile:
+				return json.load(myfile)["regeneration_done"]
+			#return os.access(os.path.join(self.source_dir, 'regenerated.simexpal'), os.F_OK)
 		return os.access(os.path.join(self.clone_dir, 'regenerated.simexpal'), os.F_OK)
 
 	def is_configured(self):
-		return os.access(os.path.join(self.compile_dir, 'configured.simexpal'), os.F_OK)
+		with open(self._cfg.basedir + '/simex.cache', 'r') as myfile:
+			return json.load(myfile)["configuration_done"]
```
c-bebop, line 1035: I see that we have to repeat these JSON keys several times. I recommend having constants defining those keys and using them as variables. This way we can't trip over misspellings.

simexpal/build.py
```diff
@@ -271,7 +272,12 @@ def git_ref_changed():
 				# Clone the git repository into the build.source_dir
 				subprocess.check_call(['git', 'clone', build.info.git_repo, build.source_dir])
 
-			util.touch(os.path.join(checkout_dir, 'checkedout.simexpal'))
+			# Mark checkout as finished
+			with open(cfg.basedir + '/simex.cache', 'r') as myfile:
+				cache = json.load(myfile)
+				cache["checkout_done"] = True
+			with open(cfg.basedir + '/simex.cache', 'w') as myfile:
```
c-bebop, line 279: Again. The rw access in two different scopes. I am wondering if there is perhaps a benefit of reading and writing in two different scopes? Is the limit in access beneficial to the separation of concern?

c-bebop (changes requested):
I have left a couple of questions, recommendations and proposals to improve. The overall approach is valid and needs no further adjustment I think.

c-bebop:
I see that there might be an issue with the current interface of GDSB. Instead of pulling the latest commit, I recommend checking out the branch `mpi_io` which is the most up to date API of GDSB.

c-bebop (approved):

PetholzA:
The latest commit amendments also adjusted the corresponding docs, thereby this LGTM.

c-bebop (approved):

PR #180, opened 2024-08-09
refactor cache

PetholzA:
this PR fixes the incorrectness occured in #177

### DIFF
--- a/docs/builds.rst
+++ b/docs/builds.rst
@@ -297,14 +297,14 @@ Below you can find the shortened directory structure of our
 `C++ example <https://github.com/hu-macsy/simexpal/tree/master/examples/sorting_cpp>`_ example. The
 repository directory has ``<build_name>`` as prefix and ``.repo`` as suffix. The clone, compilation and
 installation directory have ``<build_name>@<revision_name>`` as prefix and the first two have ``.clone``
-and ``.compile`` as suffix respectively. The installation directory does not have any suffix. The internal
-simexpal files have the suffix ``.simexpal``.
+and ``.compile`` as suffix respectively. The installation directory does not have any suffix. The internal 
+cache that memorizes the internal state is the ``.simex.cache`` file, which is found at the level of the 
+``experiments.yml`` file. 
 
 .. code-block:: bash
    :caption: Build directories for normal builds used by simexpal during the build process.
 
    /path/to/experiments.yml/directory
-   ├── CMakeLists.txt
    ├── builds
    │   ├── simexpal.repo                        # repository directory
    │   │   ├── internal simexpal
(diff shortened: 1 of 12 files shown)


docs/builds.rst
```diff
@@ -352,32 +349,28 @@ Below you can find the shortened directory structure of our
 (if ``recursive-clone`` was set to ``True``). The clone, compilation and installation directory have
 ``<build_name>@<revision_name>`` as prefix. Additionally, the compilation directory has ``.compile``
 as suffix. The clone directory is located in the ``/develop`` directory, whereas the compilation and
-installation directories are located in the ``/dev-builds`` directory. The internal simexpal files have
-the suffix ``.simexpal``.
+installation directories are located in the ``/dev-builds`` directory. The internal cache that memorizes
+the internal state is the ``.simex.cache`` file, which is found at the level of the ``experiments.yml`` file.
```
c-bebop, line 353: I think you can simply remove the last note on where to find the `.simex.cache` file since you've already mentioned that above. Therefore I recommend the change to be reduced to:
```rst
... installation directories are located in the ``/dev-builds`` directory. The internal cache that memorizes
the internal state is the ``.simex.cache`` file.
```

simexpal/build.py
```diff
@@ -65,6 +67,15 @@ class Phase(IntEnum):
 	COMPILE = 4
 	INSTALL = 5
 
+def mark_as_finished(basedir, buildname, type = None):
+	if type == None:
+		raise Exception("Wrong key, see util.CONSTANTS for valid keys")
```
c-bebop, line 72: So this exception will be raised whenever `mark_as_finished()` is called with `type = None` correct? That however is a logical programming error and not something a user would get any helpful information from whenever that would happen. I would recommend changing that to an assertion and **not** raising an exception.

tests/builds/experiments_ymls/build_examples/test_build.cpp
```diff
@@ -0,0 +1,8 @@
+#include <iostream>
+#include <vector>
+#include <fstream>
+#include <string>
+
+int main(int argc, char **argv){
```
c-bebop, line 6: I understand this is just a test file but it wouldn't hurt to use some kind of formatting option for this. Also, a c++ main function should always return an `int` as stated in it's declaration therefore I would recommend changing the code to the following (including the removal of header includes not used within this file and flushing the output stream):
```c++
#include <iostream>

int main(int argc, char **argv) {
    std::cout << "Cpp file has been build successfully." << std::endl;
    return 0;
}
```

c-bebop (changes requested):
Looks good, but I left some comments to fix minor things.

c-bebop (approved):
Thanks for spotting the GDSB MPI option!

PR #182, opened 2024-11-20
add max_simultaneous experiment setting

bernlu:
This PR adds a new parameter to the `experiments` section. This parameter limits the maximum number of runs that slurm executes at once (extending the slurm array syntax that we already use: https://slurm.schedmd.com/job_array.html).

```yml
experiments:
  - name: foo
    args: ['foo.py']
    stdout: out
    max_simultaneous: 2
```

Use case: sometimes license limitations prevent us from running more than a specific number of experiments at once.

Limitation of this implementation: since this PR just extends the slurm array syntax, this setting only applies to a single launch. When launching multiple sets of experiments, the limit will be applied to each set individually.

### DIFF
--- a/simexpal/launch/slurm.py
+++ b/simexpal/launch/slurm.py
@@ -138,7 +138,11 @@ class SlurmLauncher(common.Launcher):
 				'-e', os.path.join(cfg.basedir, 'aux/_slurm/' + log_pattern + '.err')])
 
 		if use_array:
-			sbatch_args.append('--array=0-' + str(len(locked) - 1))
+			if experiment.max_simultaneous:
+				max_simultaneous = f"%{experiment.max_simultaneous}"
+			else:
+				max_simultaneous = ""
+			sbatch_args.append(f"--array=0-{len(locked) - 1}{max_simultaneous}")
 
 		# Add custom sbatch parameters of the user.
 		slurm_args = util.ensure_list_type(experiment.effective_slurm_args)
--- a/simexpal/schemes/experiments.json
+++ b/simexpal/schemes/experiments.json
@@ -56,6 +56,10 @@
 				"slurm_args": {"$ref": "#/definitions/str_or_str_list"},
 				"exclusive": {
 					"type": "boolean"
+				},
+				"max_simultaneous": {
+					"type": "integer",
+					"minimum": 1
 				}
 			}
 		}
@@ -449,6 +453,7 @@
 							"procs_per_node": {},
 							"num_threads": {},
 							"exclusive": {},
+							"max_simultaneous": {},
 							"slurm_args": {}
 						},
 						"additionalProperties": false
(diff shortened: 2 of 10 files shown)


tests/builds/experiments_ymls/build_examples/experiments.yml
```diff
@@ -44,12 +58,12 @@ builds:
 revisions:
   - name: git-gdsb
     build_version:
-      'test-git-gdsb': '027ee5048e8c00648185ee4321dab969a80396fe' #arbitrary commit hash to ensure determinism
+      'test-git-gdsb': '3fe8b2e38d9de38b459e2e958df66577b77ccccf' #arbitrary commit hash to ensure determinism
```
c-bebop, line 61: I guess the comment is a bit... confusing. It's not determinism that we want but reproducibility.

simexpal/launch/slurm.py
```diff
@@ -138,7 +138,11 @@ def _represents_int(x):
 				'-e', os.path.join(cfg.basedir, 'aux/_slurm/' + log_pattern + '.err')])
 
 		if use_array:
-			sbatch_args.append('--array=0-' + str(len(locked) - 1))
+			if experiment.max_simultaneous:
+				max_simultaneous = f"%{experiment.max_simultaneous}"
+			else:
+				max_simultaneous = ""
+			sbatch_args.append(f"--array=0-{len(locked) - 1}{max_simultaneous}")
```
c-bebop, line 145: It would be cool to provide these arguments using a python API such as [PySlurm](https://github.com/PySlurm/pyslurm), but I see this is out of scope for this PR.

simexpal/schemes/experiments.json
```diff
@@ -56,6 +56,10 @@
 				"slurm_args": {"$ref": "#/definitions/str_or_str_list"},
 				"exclusive": {
 					"type": "boolean"
+				},
+				"max_simultaneous": {
+					"type": "integer",
+					"minimum": 1
```
c-bebop, line 62: Why not `"minimum": 0`? As in: `0` would be illegal? If the key is used we have to at least provide `0`?

tests/builds/test_builds.py
```diff
@@ -38,11 +38,31 @@ def check_cache_containment(cache_entries, dir, build_name):
...
 
     for rev_name, build_name in test_instances:
         revision = cfg.get_revision(rev_name)
         example_build = cfg.get_build(build_name, revision)
         build.make_builds(cfg, revision, [example_build.info], [build_name], wanted_phases)
-        check_cache_containment(wanted_cache_entries, example_build._cfg.basedir, build_name)    
+        check_cache_containment(wanted_cache_entries, example_build._cfg.basedir, build_name)
+
+# pytest runs this function first, then on yield runs all tests, then finally runs the code following 'yield' (as cleanup/finalize)
+@pytest.fixture(scope='session', autouse=True)
```
c-bebop, line 51: Why the cleanup? Is that for local tests? For our CI we don't need to cleanup correct?

c-bebop (dismissed):

PR #183, opened 2025-02-26
Fix job status 'broken' with slurm when job is in fact not broken

bernlu:
This issue is caused by job names with spaces in them - in that case output parsing was wrong. Fixed by just not querying for the name since it is not used at all.

### DIFF
--- a/simexpal/base.py
+++ b/simexpal/base.py
@@ -453,16 +453,17 @@ class Config:
 		if not self.slurm_queried:
 			# -h omits the header line.
 			# -r outputs one job array element per line.
-			output = subprocess.check_output(['squeue', '-h', '-r'])
+			# --format=%i,%t outputs jobid,status
+			output = subprocess.check_output(['squeue', '-h', '-r', '--format=%i,%t'])
 			output = output.decode().splitlines()
 
 			self.slurm_queried = True
 
 			if len(output) > 0:
-				output = [entry.split() for entry in output]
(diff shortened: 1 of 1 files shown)


simexpal/base.py
```diff
@@ -453,16 +453,17 @@ def query_slurm(self):
...
 
 			self.slurm_queried = True
 
 			if len(output) > 0:
-				output = [entry.split() for entry in output]
+				output = [entry.split(',') for entry in output]
 				for entry in output:
 					entry_jobid = entry[0]
-					entry_state = entry[4]
+					entry_state = entry[1]
```
c-bebop, line 466: Doesn't this require that `len(output) >= 2`?

tests/examples/slurm-pr-183/runtest.sh
```diff
@@ -0,0 +1,16 @@
+simex e purge --failed -f
```
c-bebop, line 1: Perhaps use an interpreter line here?:
```bash
#!/bin/bash
```

tests/examples/slurm-pr-183/experiments.yml
c-bebop, line 1: What's with the `tests/examples/slurm-pr-183` folder structure? Do you want to leave it as this? Or what would be the actual folder you want to put this in?

c-bebop (changes requested):
Looks good, just needs some adaptation/safety check.

